Issue #20, opened 2022-03-16
Create ERD diagram


OfirMatasas:
Define the database relations of the project

Omeramsc:
Please refer to the [issue template](https://github.com/redhat-beyond/Hotails/blob/main/Contributing.md#issue-template) when writing an issue.

ErezCohenn:
Adding a link for the online editor for the ERD diagram 
https://lucid.app/lucidspark/8969c0ba-5430-4e8c-90e1-28a23c2f3c7a/edit?invitationId=inv_63385825-f9f9-4cb2-bdac-9fb979645249#

ErezCohenn:
> Please refer to the [issue template](https://github.com/redhat-beyond/Hotails/blob/main/Contributing.md#issue-template) when writing an issue.

@OfirMatasas I don't have the permission to change it, can you change it as omer mentioned or give the permission?

PR #21, opened 2022-03-16
📈 Creating ERD diagram

ErezCohenn:
The ERD diagram represents the data model and business logic for the web application.
The database that will be built for each application will be based on that diagram.

Closing #20


OfirMatasas (changes requested):
Great job overall. Few things to take into consideration:
1. Delete the status options in the Order table, as we don't have this kind of information in any other field over the diagram.
2. Add variable types in all fields, for better synchronization between the models, when working parallelly. For example, in Dog Owner, the dog name can be TextField(20), and Dog age can be Integer.
3. The logo added to the ERD is not the logo of our project. Either change it to the correct one or remove it.

ormergi:
Could you please explain what is the "User" entity is used for?

ormergi:
Hi Erez,
Could you please give some context of who is the `User` entity?
As far as I understand there is already an entity to represent a user which is `DogOwner`,
why is it necessary to have another one?

ErezCohenn:
> Great job overall. Few things to take into consideration:
> 
> 1. Delete the status options in the Order table, as we don't have this kind of information in any other field over the diagram.
> 2. Add variable types in all fields, for better synchronization between the models, when working parallelly. For example, in Dog Owner, the dog name can be TextField(20), and Dog age can be Integer.
> 3. The logo added to the ERD is not the logo of our project. Either change it to the correct one or remove it.

1. I think that the field `Status` is not recognized in any other table is not a reason to delete it. I think that this is a necessary field for order reservation. tell me what u think team and we will decide about it together
2. Agree, I will add that soon.
3. Good cathing, will be changed either.

OfirMatasas:
> > Great job overall. Few things to take into consideration:
> > 
> > 1. Delete the status options in the Order table, as we don't have this kind of information in any other field over the diagram.
> > 2. Add variable types in all fields, for better synchronization between the models, when working parallelly. For example, in Dog Owner, the dog name can be TextField(20), and Dog age can be Integer.
> > 3. The logo added to the ERD is not the logo of our project. Either change it to the correct one or remove it.
> 
> 1. I think that the field `Status` is not recognized in any other table is not a reason to delete it. I think that this is a necessary field for order reservation. tell me what u think team and we will decide about it together
> 2. Agree, I will add that soon.
> 3. Good cathing, will be changed either.

About 1, you got a point. Either keeping it or deleting it works for me.

OfirMatasas (approved):

tamirmatok (changes requested):
> Hi Erez, Could you please give some context of who is the `User` entity? As far as I understand there is already an entity to represent a user which is `DogOwner`, why is it necessary to have another one?

Agree with that . I think it would be better to delete 'User' entity and modify 'DayCare' and 'DogOwner' as needed .

ErezCohenn:
> modify

I agree, but I think we are still want to use the `User` entity that is build-in Django that will make it easier for us to manage users.
 Someone knows if there is a way we can use the `User` entity and insert fields to him so that at the end we will stay with 2 entities: `DayCare` and `DogOwner` that have common fields that exist in a `User` entity instead of 3 entities.

jeniaSakirko (commented):
@ErezCohenn can you add the file so we could work on it together?

Omeramsc (changes requested):
- I think you can keep `User`, but it shouldn't include the first\last name or phone number. if those are relevant, they can be placed in the entity itself.
- separate the line from user to `Dog Owner` and `DayCare`. it is confusing, as those entities are not related to each other.
- nit: s/DayCare/Day Care, or remove the spaces in all enteties.
- What does `PropertyID` under `ReviewBoard` refers to? is should be clear. I also don't think `ReviewBoard` should be an entity at all, it is a feature that is using the `review` entity.
- Same goes for `Message Boards`

ErezCohenn:
> @ErezCohenn can you add the file so we could work on it together?
I added the online editor link to the issue so anyone will be able to edit the diagram.
Take a look here: #20

Omeramsc:
what happens when a Dogowner has more than one dog?

ErezCohenn:
> * I think you can keep `User`, but it shouldn't include the first\last name or phone number. if those are relevant, they can be placed in the entity itself.
> * separate the line from user to `Dog Owner` and `DayCare`. it is confusing, as those entities are not related to each other.
> * nit: s/DayCare/Day Care, or remove the spaces in all entities.
> * What does `PropertyID` under `ReviewBoard` refers to? it should be clear.
> * I don't think `ReviewBoard` should be an entity at all, it is a feature that is using the `review` entity.
> * Same goes for `Message Boards`

The ERD diagram fixed and changed

Yuval-Vino (approved):
LGTM , time to implement!

tamirmatok (approved):
Great Job ! @ErezCohenn

ormergi:
@ErezCohenn If I understand correctly User is the builtin django entity, right? if so please add correspond note.

ErezCohenn:
> @ErezCohenn If I understand correctly User is the builtin django entity, right? if so please add correspond note.

You're right, just added.

Omeramsc (commented):
- Why do we have `Review_ID` twice under Review?
- Same goes for `Message_ID`
- How do you plan `message` to work? is it a single private message, or a chat? I find the `Author` field redundant, and confusing (it is not connected to any table, how will it identify?)

Other than that - very good work!!

ErezCohenn:
> * Why do we have `Review_ID` twice under Review?
> * Same goes for `Message_ID`
> * How do you plan `message` to work? is it a single private message, or a chat? I find the `Author` field redundant, and confusing (it is not connected to any table, how will it identify?)
> 
> Other than that - very good work!!

Notice that the `Review_ID int (FK, self)` means that its reference to another entity of review, so the review can chain,
The same is working for messages.

Omeramsc:
> > * Why do we have `Review_ID` twice under Review?
> > * Same goes for `Message_ID`
> > * How do you plan `message` to work? is it a single private message, or a chat? I find the `Author` field redundant, and confusing (it is not connected to any table, how will it identify?)
> > 
> > Other than that - very good work!!
> 
> Notice that the `Review_ID int (FK, self)` means that its reference to another entity of review, so the review can chain, The same is working for messages.

Why would you want to chain them? 
if you want to order them one after the other on the same page, you can use the datetime to do it. 
if you want a reply to a review, it can be normal varchar (we already know who is replying)

ErezCohenn:
> > > * Why do we have `Review_ID` twice under Review?
> > > * Same goes for `Message_ID`
> > > * How do you plan `message` to work? is it a single private message, or a chat? I find the `Author` field redundant, and confusing (it is not connected to any table, how will it identify?)
> > > 
> > > Other than that - very good work!!
> > 
> > 
> > Notice that the `Review_ID int (FK, self)` means that its reference to another entity of review, so the review can chain, The same is working for messages.
> 
> Why would you want to chain them? if you want to order them one after the other on the same page, you can use the datetime to do it. if you want a reply to a review, it can be normal varchar (we already know who is replying)
> 
> Anyway, you can't have two of the same key on a table.

But in a situation where we want a reply upon reply, how can this be realized in the method you have proposed? (by adding normal varchar field), I think we must chain the messages as we learned in the class.

Omeramsc:
> 

In most popular review systems (Facebook, Google, AliExpress, etc), a review usually consist of 1 review and 1 reply, not a conversation. I don't see a reason for you to complicate things for yourself.

Omeramsc:
> > > > * Why do we have `Review_ID` twice under Review?
> > > > * Same goes for `Message_ID`
> > > > * How do you plan `message` to work? is it a single private message, or a chat? I find the `Author` field redundant, and confusing (it is not connected to any table, how will it identify?)
> > > > 
> > > > Other than that - very good work!!
> > > 
> > > 
> > > Notice that the `Review_ID int (FK, self)` means that its reference to another entity of review, so the review can chain, The same is working for messages.
> > 
> > 
> > Why would you want to chain them? if you want to order them one after the other on the same page, you can use the datetime to do it. if you want a reply to a review, it can be normal varchar (we already know who is replying)
> > Anyway, you can't have two of the same key on a table.
> 
> But in a situation where we want a reply upon reply, how can this be realized in the method you have proposed? (by adding normal varchar field), I think we must chain the messages as we learned in the class.

In most popular review systems (Facebook, Google, AliExpress, etc), a review usually consist of 1 review and 1 reply, not a conversation. I don't see a reason for you to complicate things for yourself.

ErezCohenn:
> > > > > * Why do we have `Review_ID` twice under Review?
> > > > > * Same goes for `Message_ID`
> > > > > * How do you plan `message` to work? is it a single private message, or a chat? I find the `Author` field redundant, and confusing (it is not connected to any table, how will it identify?)
> > > > > 
> > > > > Other than that - very good work!!
> > > > 
> > > > 
> > > > Notice that the `Review_ID int (FK, self)` means that its reference to another entity of review, so the review can chain, The same is working for messages.
> > > 
> > > 
> > > Why would you want to chain them? if you want to order them one after the other on the same page, you can use the datetime to do it. if you want a reply to a review, it can be normal varchar (we already know who is replying)
> > > Anyway, you can't have two of the same key on a table.
> > 
> > 
> > But in a situation where we want a reply upon reply, how can this be realized in the method you have proposed? (by adding normal varchar field), I think we must chain the messages as we learned in the class.
> 
> In most popular review systems (Facebook, Google, AliExpress, etc), a review usually consist of 1 review and 1 reply, not a conversation. I don't see a reason for you to complicate things for yourself.

Got your point, I will update the reviews table.
What about the messages? they should be a chat and not s single private message, so I think we have to stay with the
 `Message_ID int (FK, self)` field, what u think?

tamirmatok:
I have notice that Django "User" entity includes non-mendatory fields of "First name" and "Last name",
should we still keep those fields in "Dog owner" fields ? @Omeramsc

Omeramsc:
@ErezCohenn You can try using it with messages if that works for you (change the name of the key though).

@tamirmatok You tell me - does every entity that extends `User` need a first and last name properties?

ErezCohenn:
> @ErezCohenn You can try using it with messages if that works for you (change the name of the key though).

Done

tamirmatok:
> @ErezCohenn You can try using it with messages if that works for you (change the name of the key though).
> 
> @tamirmatok You tell me - does every entity that extends `User` need a first and last name properties?

I think that we can give up on them @ErezCohenn  @Omeramsc

Omeramsc (approved):
LGTM

ormergi (approved):
Thanks

PR #28, opened 2022-03-19
adding review data model

### LINKED ISSUE #27: Create review database and app
Create a database and app to store the reviews from dog owners about the daycares. 
Each review should have unique id, the review content, rating and references to the dog owner and the daycare.

jeniaSakirko:
Adding review model with basic information.
Adding review to the settings.py

Depends on #41  
Resolves #27 

Signed-off-by: Jenia Sakirko <jenia.sakirko@gmail.com>


review/models.py
```diff
@@ -0,0 +1,9 @@
+from django.db import models
+
+
+class Review(models.Model):
+    review = models.TextField()
+    # reviewer_id = models.ForeignKey('user_id', on_delete = models.DO_NOTHING)
```
Omeramsc, line 6: Is the reviewer a DogOwner?

review/models.py
```diff
@@ -0,0 +1,9 @@
+from django.db import models
+
+
+class Review(models.Model):
```
Omeramsc, line 4: I'm missing a PK here

review/models.py
```diff
@@ -0,0 +1,9 @@
...
+    review = models.TextField()
+    # reviewer_id = models.ForeignKey('user_id', on_delete = models.DO_NOTHING)
+    # daycare_id = models.ForeignKey('daycare', on_delete = models.CASCADE)
+    rating = models.IntegerField(default=0)
```
Omeramsc, line 8: Isn't there a min\max value?

Omeramsc (changes requested):
Do note this won't be merged without tests.

Also, since it depends heavily on having a DogOwner\DayCare entity- it should be marked as a draft\WIP for the time being.

jeniaSakirko:
> Do note this won't be merged without tests.
> 
> Also, since it depends heavily on having a DogOwner\DayCare entity- it should be marked as a draft\WIP for the time being.

you are right, changed it

review/models.py
jeniaSakirko: yes

review/models.py
jeniaSakirko: the PK is the default generated ID

review/models.py
Omeramsc: then it should be mentioned or renamed.

review/models.py
Omeramsc: You're right, disregard.

ormergi:
ping

ErezCohenn (commented):
GJ @jeniaSakirko

review/models.py
```diff
@@ -0,0 +1,20 @@
...
+    dogowner_id = models.ForeignKey(DogOwner, on_delete=models.DO_NOTHING)
+    daycare_id = models.ForeignKey(DayCare, on_delete=models.CASCADE)
+    rating = models.IntegerField(validators=[MinValueValidator(0), MaxValueValidator(5)])
+    creation_date = models.DateTimeField(auto_now_add=True)
+
+
+@receiver(post_save, sender=Review)
+def save_user_profile(sender, instance, **kwargs):
+    if instance.full_clean():
+        instance.save()
```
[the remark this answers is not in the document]
jeniaSakirko: I think signals is more clean solution here, when you use the model it will be like nothing was changed

review/tests.py
```diff
@@ -0,0 +1,46 @@
...
+        with pytest.raises(ValidationError):
+            tmp_review = Review(review=REVIEW, rating=-1, daycare_id=DayCare.objects.get(pk=DAY_CARE_ID),
+                                dogowner_id=DogOwner.objects.get(pk=DOG_OWNER_ID))
+            tmp_review.save()
+
+    def test_rating_limit_max(self):
+        with pytest.raises(ValidationError):
+            tmp_review = Review(review=REVIEW, rating=6, daycare_id=DayCare.objects.get(pk=DAY_CARE_ID),
+                                dogowner_id=DogOwner.objects.get(pk=DOG_OWNER_ID))
+            tmp_review.save()
```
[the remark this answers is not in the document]
jeniaSakirko: I'll add a test case like you suggested :) 
You can see [here](https://github.com/redhat-beyond/Hotails/pull/28/files#diff-d70aa92e7637c74f52bf803876cd4aac2eca656bd73fb6d640b61e8f5f59ba66R11-R12) that the review is deleted only when DayCare deleted

review/models.py
jeniaSakirko: Yes, it was updated with validations

tamirmatok (approved):

review/models.py
```diff
@@ -0,0 +1,20 @@
...
+class Review(models.Model):
+    review = models.TextField()
+    dogowner_id = models.ForeignKey(DogOwner, on_delete=models.DO_NOTHING)
+    daycare_id = models.ForeignKey(DayCare, on_delete=models.CASCADE)
+    rating = models.IntegerField(validators=[MinValueValidator(0), MaxValueValidator(5)])
+    creation_date = models.DateTimeField(auto_now_add=True)
+
+
+@receiver(post_save, sender=Review)
+def save_user_profile(sender, instance, **kwargs):
```
ormergi, line 18: Why this function is necessary?
You could just call save() where needed, isnt it?

review/models.py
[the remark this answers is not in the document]
ormergi: A more simple solution would be to define static create method that also call save(),
it will also simplify the tests, and consistent with other models.

In general I find using signals just for persisting an object misuse.
Moreover I saw that save() is called on tests, if it doesn't work w/o it 
I am not sure  what is the value of having `save_user_profile`.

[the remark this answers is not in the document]
jeniaSakirko: The purpose of this function is to run the full clean (e.g. validators that were defined) and save the object it it passed.
In this case we just run the regular save() and the signals calling this function and running the validation.
We can also create static function that we will the same but then we need to make sure to use it instead of the save directly and in case someone will call save directly the object will be created without the validations

review/models.py
jeniaSakirko: See comment bellow, we are not calling this function directly its just making sure that each save is validated

review/models.py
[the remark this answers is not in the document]
ormergi: Validations should be part of the object creation flow, if the arguments are not valid the object should not be created.
In this case you can define static create method / constructor with the validation and persistence logic.
This way it is guaranteed that a created object is valid and persisted, w/o using signal magics.

ormergi (changes requested):

ErezCohenn (approved):

review/tests.py
```diff
@@ -0,0 +1,61 @@
+import pytest
+from .models import Review, DayCare, DogOwner
+from django.core.validators import ValidationError
+import logging
+
+LOGGER = logging.getLogger(__name__)
```
ormergi, line 6: It seems that it not begin used

review/tests.py
```diff
@@ -0,0 +1,61 @@
...
+
+@pytest.mark.django_db
+class TestReviewModel:
+    def test_persist_review(self, test_review):
+        assert test_review.review == REVIEW
+        assert test_review.rating == RATING
+        assert test_review.daycare_id == DayCare.objects.get(pk=DAY_CARE_ID)
+        assert test_review.daycare_id_id == DAY_CARE_ID
+        assert test_review.dogowner_id == DogOwner.objects.get(pk=DOG_OWNER_ID)
+        assert test_review.dogowner_id_id == DOG_OWNER_ID
```
ormergi, line 28: There is a problem with asserting each class member, if tomorrow the model is changed
this test should be changed as well, which is kind of missing the point in having it.
Also, note that there is no actual logic to check here, its just assigning arguments to the class members.

ormergi (commented):
Overall it looks good, see my in line comments

review/tests.py
jeniaSakirko: You are right, fixed

review/tests.py
jeniaSakirko: removed the assignment tests, in case we will have logic in the create we will add specific tests for that

ormergi (approved):
Thanks you, it looks good.

OfirMatasas (approved):
LGTM

ormergi:
@jeniaSakirko  You need to rebase

jeniaSakirko:
> @jeniaSakirko You need to rebase

@ormergi Done

ormergi (approved):
Thank you it looks good!

PR #30, opened 2022-03-20
Adding PyTest to project

### LINKED ISSUE #29: Add PyTest to workflows
Just like Flake8, add PyTest to our repo checks, so any code we add to the project, following tests, will get tested before merging.

Must include:
- pytest.imi, including tests.py, test_*.py, *_tests.py type of files
- pytest.yml, for running PyTest
- A "truth" test, so show that PyTest works

OfirMatasas:
Like Flake8, PyTest will check all tests and make sure there are no
faults in added code.

closes #29


.github/workflows/pytest.yml
```diff
@@ -0,0 +1,20 @@
...
+  pytest:
+    name: Run the application's tests with PyTest
+    runs-on: ubuntu-20.04
+    container: fedora:34
+    steps:
+      - name: Install Pipenv and git
+        run: dnf install -y pipenv git
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup Python
```
Omeramsc, line 13: What is this step needed for?

Omeramsc (changes requested):

.github/workflows/pytest.yml
OfirMatasas: Tried few actions to make it work.
Do you think this is the cause of the check failure?

ErezCohenn (commented):
@OfirMatasas  Don't forget to assign yourself for the PR and add signoff to the commits as  we wrote in the `Contributing` file

Pipfile
```diff
@@ -8,6 +8,7 @@ django = "*"
 
 [dev-packages]
 flake8 = "*"
+pytest-django = "*"
```
ErezCohenn, line 11: Try to add the line: `pytest = "*"` in addition to `pytest-django = "*"`.
Also, I think maybe you need to re-create the Pipfile.lock as @Omeramsc mentioned before because you edit the file Pipfile.

Omeramsc (approved):
LGTM, 
just make sure to reset the pipfile.lock when changing it.

ErezCohenn (approved):
Great job @OfirMatasas !

tamirmatok (approved):

.github/workflows/pytest.yml
OfirMatasas: It's from yesterday's presentation about installing PyTest in our project.
image

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,20 @@
...
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+      - name: Setup environment
+        run: pipenv sync --dev
+      - name: Run PyTest
+        run: pipenv run pytest -v
```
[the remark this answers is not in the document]
OfirMatasas: Apparently it's not important to this kind of files.
Just like Flake8, it works without a new line in the end of the file.

.github/workflows/pytest.yml
Omeramsc: I'm asking about `name: Setup Python` specifically. I don't see it in the presentation either.

OfirMatasas: > I'm asking about `name: Setup Python` specifically. I don't see it in the presentation either.
You're right. Added it to make things work after few failures.
I'll remove it, as I think it doesn't make any difference.

.github/workflows/pytest.yml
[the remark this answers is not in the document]
ormergi: Note there is a long discussion about adding EOL (newline character) at EOF on various platforms,
not adding new line at the end of a file may result in unexpected behavior of different tools (e.g: ignoring the last line of the file).
Also most text editors will add it automatically, so next time this file is edited it will be part of the diff.
Moreover it is not consistent with most of the files in the project.

Pipfile
```diff
@@ -5,9 +5,12 @@ name = "pypi"
...
+pytest = "*"
 
 [dev-packages]
 flake8 = "*"
+pytest-django = "*"
+pytest = "*"
 
 [requires]
-python_version = "3.9"
+python_version = "3.9"
```
ormergi, line 16: Please add missing newline

ormergi (changes requested):
Hi Ofir, please see my inline comments.

ormergi (approved):
Thanks

PR #32, opened 2022-03-20
🔧  Fixing the provisioned test server that will not crash

### LINKED ISSUE #31: Fix: Make provisioned test server not crash
The way we launched the Django dev server in the provision script had
caused it to crash every time we change the code.
 With the new command, it reloads the code as it needs to.
- [ ] Fix the runserver command in setup.sh file

ErezCohenn:
The way we launched the Django dev server in the provision script had
caused it to crash every time we change the code.
With the new command that updated on setup.sh file, it reloads the web application as we will edit the code

Close #31 

Signed-off-by: Erez Cohen <erez6356458@gmail.com>

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -10,6 +10,7 @@ cd /vagrant
 # Install dependencies with Pipenv
 pipenv sync --dev
 
-# run our app. Nohup and "&" are used to let the setup script finish
-# while our app stays up. The app logs will be collected in nohup.out
-nohup pipenv run python manage.py runserver 0.0.0.0:8000 &
+# run our app. setsit, the parentheses and "&" are used to perform a "double
+# fork" so that out app stays up after the setup script finishes.
+# The app logs are redirected to the `runserver.log` file.
+(setsid pipenv run python manage.py runserver 0.0.0.0:8000 > runserver.log 2>&1 &)


tamirmatok (approved):
Good job @ErezCohenn

setup.sh
```diff
@@ -10,6 +10,7 @@ cd /vagrant
 # Install dependencies with Pipenv
 pipenv sync --dev
 
-# run our app. Nohup and "&" are used to let the setup script finish
-# while our app stays up. The app logs will be collected in nohup.out
-nohup pipenv run python manage.py runserver 0.0.0.0:8000 &
+# run our app. setsit, the parentheses and "&" are used to perform a "double
+# fork" so that out app stays up after the setup script finishes.
+# The app logs are redirected to the `runserver.log` file.
+(setsid pipenv run python manage.py runserver 0.0.0.0:8000 > runserver.log 2>&1 &)
```
Omeramsc, line 16: what made the app crash with what you used before? I never saw it happen.

ErezCohenn: Always when we changed the code of the project the server was failed and we had to run the command:
 `vagrant provision`. 
Now it should be resolved as Barak Korren mentioned in the slack [here](https://redhat-beyond.slack.com/archives/C031MUZG0LE/p1647795038129999)

Omeramsc: Right, thanks. :)

PR #35, opened 2022-03-22
Adding .DS_Store to .gitignore

### LINKED ISSUE #34: Add .DS_Store to gitignore
As developer I don't want to see .DS_Store when running `git status` and I don't want to commit the file

AC:
.DS_Store is not pushed to the repo
.DS_Store is not visible when running `git status`

jeniaSakirko:
.DS_Store is MacOS file thats stores custom attributes of its containing folder https://en.wikipedia.org/wiki/.DS_Store.
Such file should not be committed 

closes #34 

Signed-off-by: Jenia Sakirko <jenia.sakirko@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Mac OS
+**/.DS_Store
+
 # IDEA
 .idea
 


ErezCohenn (commented):
Can you add some description for the PR why we have to add that type of file to the gitignore.
Also dont forget to add sign-off to the PR

PR #41, opened 2022-03-22
⚔️ Creating Daycare app

ErezCohenn:
Preparation of an initial database for the daycare application,
where a daycare owner will be able to register to the site.
In addition, the daycare owner will be able to manage his site and edit his profile in the future.
 
how is done:
- [x] Adding DayCare app to the project that includes: bootstrapping files, DB model class and migrations
- [x] Updating settings.py file in order to use the DayCare app
- [x] Updating .gitignore file in order to ignore db.sqlite3 file
- [x] Testing the app files

Closing #24

PR #33, opened 2022-03-21
Adding Dog-Owner appliction 

tamirmatok:
Adding Dog-owner application. 
Features implemented: 
- creation of new dog-owner and user.

Dog-owner is exstending Django 'User' model .

Including: tests, migrations file, and data models .

closes #26

PR #46, opened 2022-03-26
Creating orders app

### LINKED ISSUE #22: Create orders database and app
Create an order database and app to store the history and future orders between dog owners and daycares, filling the rules described in the project ERD diagram.

OfirMatasas:
The order app manage the reservations between the dog owner and the daycare (including tests).

Possible actions in this app:
- Create new order as a dog owner
- Accept/decline an order as a daycare
- Cancel an order as both daycare and dog owner
- Watch order history (only showing orders that both the daycare and the dog owner of it were not deleted)

Depends on #33 #41
Closes #22 


orders/models.py
```diff
@@ -0,0 +1,84 @@
...
+    Canceled = 'C', 'Canceled'
+    Finished = 'F', 'Finished'
+    OnGoing = 'O', 'On Going'
+
+
+class Order(models.Model):
+    order_id = models.AutoField(primary_key=True)
+    dog_owner_id = models.IntegerField()
+    daycare_id = models.IntegerField()
+    booking_date = models.DateTimeField(auto_now=True)
```
ErezCohenn, line 17: I think that the fields: `dog_owner_id` and `daycare_id` should be mentioned as FK

orders/tests.py
```diff
@@ -0,0 +1,83 @@
...
+
+        result = order.update_status(new_status=new_status)
+        assert result != "invalid"
+        order.refresh_from_db()
+        print(f"Canceling date: {order.canceling_date.__format__('%Y-%m-%d %H:%M:%S')}")
+
+    def test_order_migrate(self):
+        order = Order.objects.get(dog_owner_id='12345')
+
+        assert order in Order.objects.all()
```
ErezCohenn, line 83: consider adding the tests:
- status max length (you can use the validators built-in Django as I used in the daycare app
- test what happens if we delete daycare/dogowner user - I think the order should remain anyway in order to see the history of orders anyway

ErezCohenn (commented):
Great job @OfirMatasas  the app looks very impressive
one more thing: I think you need to squash the commits because the test depends on the order app

orders/models.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class Order(models.Model):
+    order_id = models.AutoField(primary_key=True)
+    dog_owner_id = models.ForeignKey(DogOwner, on_delete=models.DO_NOTHING)
+    daycare_id = models.ForeignKey(DayCare, on_delete=models.DO_NOTHING)
+    booking_date = models.DateTimeField(auto_now=True)
+    approving_date = models.DateTimeField(blank=True, null=True)
+    starting_date = models.DateTimeField()
+    finishing_date = models.DateTimeField()
```
jeniaSakirko, line 22: I think better approach here should be saving `duration` instead of `finishing_date` and the date will be calculated from the `starting_date` + `duration` for exaple:
```
import datetime as DT
today = DT.date.today()
week_ago = today + DT.timedelta(days=duration)
```

orders/tests.py
```diff
@@ -0,0 +1,99 @@
...
+
+STARTING_DATE = timezone.now()
+FINISHING_DATE = timezone.now()
+TOTAL_PRICE = 100
+STATUS = StatusOptions.Pending
+
+
+@pytest.fixture
+def order():
+    return Order(dog_owner_id=DogOwner.create(email='test_dog_owner556@gmail.com',
```
jeniaSakirko, line 16: why not using DogOwner and DayCare that were created in the data migrations? in the current situation you might have conflicts with the data that was created there

orders/models.py
```diff
@@ -0,0 +1,86 @@
...
+    status = models.CharField(max_length=1,
+                              choices=StatusOptions.choices,
+                              default='P'
+                              )
+
+    def __str__(self):
+        return f"Dog owner ID: {self.dog_owner_id}, daycare ID: {self.daycare_id}, status: {self.status}"
+
+    @staticmethod
+    def create(dog_owner_id, daycare_id, starting_date, finishing_date, total_price):
```
jeniaSakirko, line 34: why do we need the create function?

orders/models.py
OfirMatasas: > I think that the fields: `dog_owner_id` and `daycare_id` should be mentioned as FK


They weren't FK until the merge of DogOwner model.
Changed it afterwards.

orders/models.py
OfirMatasas: > I think better approach here should be saving `duration` instead of `finishing_date` and the date will be calculated from the `starting_date` + `duration` for exaple:
> 
> ```
> import datetime as DT
> today = DT.date.today()
> week_ago = today + DT.timedelta(days=duration)
> ```

The idea comes from the need of importing those dates into the dog owner's and daycare's calendars, and because the dog owner chooses the starting and finishing date of an order, just like Airbnb or Booking.
So converting it from DateTime to duration, and then from duration back to DateTime feels like a waste.
Are you agree?

orders/models.py
OfirMatasas: > why do we need the create function?

The dog owner chooses the dates of a requested order, and when he decides to place the order, the order needs to be created and then saved in the database with the fields mentioned.
Do you think we should try something different?

orders/tests.py
OfirMatasas: > why not using DogOwner and DayCare that were created in the data migrations? in the current situation you might have conflicts with the data that was created there

Good idea. I'll change it.

orders/models.py
jeniaSakirko: In the calendars scenario I agree to put the data in calendar form its easier to have two DateTimes but what about reports? for example calculating the total price is much easier with duration

I think that there are pros and cons for both ways let's see what others think

orders/tests.py
```diff
@@ -0,0 +1,79 @@
...
+                                       finishing_date=order.finishing_date,
+                                       total_price=order.total_price)
+
+        order_to_delete.delete()
+        assert order_to_delete not in Order.objects.all()
+
+    def test_get_all_orders_by_dog_owner_id(self, order):
+        dog_owner_orders = Order.get_all_dog_owner_orders(order.dog_owner_id)
+
+        print(dog_owner_orders)
```
Yuval-Vino, line 47: I don't understand what are we testing here, 
It seems like you only print something, maybe we print wrong data?

orders/tests.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def test_get_all_orders_by_daycare_id(self, order):
+        order.save()
```
Yuval-Vino, line 50: Is there a reason to save the order?

orders/tests.py
```diff
@@ -0,0 +1,79 @@
...
+
+        order_to_delete.delete()
+        assert order_to_delete not in Order.objects.all()
+
+    def test_get_all_orders_by_dog_owner_id(self, order):
+        dog_owner_orders = Order.get_all_dog_owner_orders(order.dog_owner_id)
+
+        print(dog_owner_orders)
+
+    def test_get_all_orders_by_daycare_id(self, order):
```
Yuval-Vino, line 49: Same comment as above

pytest.ini
```diff
@@ -1,4 +1,4 @@
 [pytest]
 DJANGO_SETTINGS_MODULE = Hotails.settings
 django_debug_mode = true
-python_files = tests.py test_*.py *_tests.py
+python_files = tests.py test_*.py *_tests.py *_test.py
```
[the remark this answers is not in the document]
ErezCohenn: I think one of them ends with 's':
*_tests and the other is *_test

orders/models.py
OfirMatasas: Exactly what I was thinking.
Plus, we can calculate the total price per night, so it would be easier

orders/tests.py
OfirMatasas: You're right, it tests that the function do pull dog owner's orders from the database.
Maybe the check should be different, like asking for each order if it's in the database or something.
I'll change it.

orders/tests.py
OfirMatasas: My bad. I'll remove it.

orders/models.py
```diff
@@ -0,0 +1,76 @@
...
+class StatusOptions(models.TextChoices):
+    Pending = 'P', 'Pending'
+    Approved = 'A', 'Approved'
+    Canceled = 'C', 'Canceled'
+    Finished = 'F', 'Finished'
+    OnGoing = 'O', 'On Going'
+
+
+class Order(models.Model):
+    order_id = models.AutoField(primary_key=True)
```
jeniaSakirko, line 16: Why not using the default id?

orders/tests.py
```diff
@@ -0,0 +1,79 @@
...
+STARTING_DATE = timezone.now()
+FINISHING_DATE = timezone.now()
+TOTAL_PRICE = 100
+STATUS = StatusOptions.Pending
+
+
+@pytest.fixture
+def order():
+    return Order.create(dog_owner_id=DogOwner.objects.get(pk=1),
+                        daycare_id=DayCare.objects.get(pk=1),
```
jeniaSakirko, line 18: You should use `objects.get(id=...)` instead of `objects.get(pk=...)`, in case the pk will be changed this code won't be affected and also you are looking by id so its better to be explicit

orders/models.py
```diff
@@ -0,0 +1,76 @@
...
+    def update_status(self, new_status):
+        old_status = self.status
+
+        if new_status == StatusOptions.Canceled and old_status != StatusOptions.Canceled:
+            self.canceling_date = timezone.now()
+        elif new_status == StatusOptions.Approved and old_status != StatusOptions.Approved:
+            self.approving_date = timezone.now()
+
+        self.status = new_status
+        self.save()
```
ErezCohenn, line 76: This method is only updating the status to Cancel/Approve status?
I think that all the updates of status methods updating should consist of more logic and we need to separate each method for each update status, for example:
`def cancel_order` if an order is canceling (it can be the dogowner canceling the order or the daycare), then we need to update the status to Canceled. 
tell me what u think 
**note**: i think we can implement those methods in the future on another PR

orders/tests.py
```diff
@@ -0,0 +1,79 @@
...
+    def test_delete_order(self, order):
+        assert order in Order.objects.all()
+        order.delete()
+        assert order not in Order.objects.all()
+
+    def test_get_all_orders_by_dog_owner_id(self, order):
+        dog_owner_orders = Order.get_all_dog_owner_orders(order.dog_owner_id)
+
+        for dog_owner_order in dog_owner_orders:
+            assert dog_owner_order in Order.objects.all()
```
jeniaSakirko, line 44: What are you testing here? if I understood right you got a sub list of all the objects and then iterated over the list and verified that all values that you got exist in the big list? if so this looks like redundant test

orders/tests.py
```diff
@@ -0,0 +1,79 @@
...
+                        daycare_id=DayCare.objects.get(pk=1),
+                        starting_date=STARTING_DATE,
+                        finishing_date=FINISHING_DATE,
+                        total_price=TOTAL_PRICE,
+                        )
+
+
+@pytest.mark.django_db
+class TestOrderModel:
+    def test_create_order(self, order):
```
jeniaSakirko, line 27: the `order` value is not used

orders/tests.py
```diff
@@ -0,0 +1,79 @@
...
+        order.delete()
+        assert order not in Order.objects.all()
+
+    def test_get_all_orders_by_dog_owner_id(self, order):
+        dog_owner_orders = Order.get_all_dog_owner_orders(order.dog_owner_id)
+
+        for dog_owner_order in dog_owner_orders:
+            assert dog_owner_order in Order.objects.all()
+
+    def test_get_all_orders_by_daycare_id(self, order):
```
jeniaSakirko, line 46: same as `test_get_all_orders_by_dog_owner_id`

orders/tests.py
```diff
@@ -0,0 +1,79 @@
...
+        for daycare_order in daycare_orders:
+            assert daycare_order in Order.objects.all()
+
+    def test_get_order_by_order_id(self, order):
+        order = Order.objects.filter(order_id=order.order_id).first()
+        assert order is not None
+
+    def test_get_not_existed_order_by_order_id(self):
+        try:
+            Order.objects.get(order_id=-1)
```
jeniaSakirko, line 58: order_id 1 is created in the data migration no? if not the test will be effected by the data migration
Maybe create new order and than you will know the last id that was created

jeniaSakirko (changes requested):

orders/models.py
OfirMatasas: Good idea. Changing it now.

OfirMatasas: Good idea. Changing it now.

orders/models.py
OfirMatasas: > This method is only updating the status to Cancel/Approve status? 

Yes, because those are the only 2 statuses that are possible for updating order status.

> I think that all the updates of status methods updating should consist of more logic and we need to separate each method for each update status, for example: `def cancel_order` if an order is canceling (it can be the dogowner canceling the order or the daycare), then we need to update the status to Canceled. 

Maybe you're right, but I don't know if it's such a big difference.
Of course, changing the status of the order as a command will be easier, so maybe your way is better.
I'll think about it. Thank you!

orders/tests.py
OfirMatasas: You're right again. I'll change it.

orders/tests.py
OfirMatasas: > I guess the better test will be to check if `order` is in `dog_owner_orders`

I've tried to do so, but it didn't work as expected.
But it is kind of a redundant test. I'll pass that.

orders/tests.py
OfirMatasas: > order_id 1 is created in the data migration no? if not the test will be effected by the data migration Maybe create new order and than you will know the last id that was created

The test is asking for order_id = -1, which is not possible. That's why the "get" method will throw an exception, and the test will pass.

orders/tests.py
```diff
@@ -0,0 +1,79 @@
...
+            assert False
+        except ObjectDoesNotExist:
+            assert True
+
+    def test_approve_order(self, order):
+        order.save()
```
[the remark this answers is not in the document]
OfirMatasas: > the save() is part of the create function so why do you need to run it again?

It was part of the old implementation. Removed.

orders/models.py
```diff
@@ -0,0 +1,76 @@
...
+    Approved = 'A', 'Approved'
+    Canceled = 'C', 'Canceled'
+    Finished = 'F', 'Finished'
+    OnGoing = 'O', 'On Going'
+
+
+class Order(models.Model):
+    order_id = models.AutoField(primary_key=True)
+    dog_owner_id = models.ForeignKey(DogOwner, on_delete=models.CASCADE)
+    daycare_id = models.ForeignKey(DayCare, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
OfirMatasas: > I think that on_delete should be models.do_nothing. As DogOwner I wish to see all orders that I did even if the DayCare was deleted As DayCare I wish to see all orders that I had even if the DogOwner deleted.
> 
> We should delete the order only if both deleted

The "DO_NOTHING" will not work, because when a user (dog owner / daycare) will be deleted from the database, the FK will refer to a non-existed instance, and it's not possible.
I'll change it to "SET_NULL" instead.

Yuval-Vino (approved):
LGTM, Great work @OfirMatasas !

jeniaSakirko (approved):

orders/models.py
```diff
@@ -0,0 +1,79 @@
...
+    def get_all_daycare_orders(user_id):
+        filter_args = {
+            'daycare_id': user_id,
+        }
+
+        return Order.objects.filter(**filter_args)
+
+    @staticmethod
+    def get_order_by_order_id(order_id):
+        return Order.objects.filter(order_id=order_id.id)
```
ormergi, line 64: Please try to keep the code more consistent, you can pass filter args directly similar to get_order_by_order_id().

orders/models.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def approve_order(self):
+        if self.status == StatusOptions.Pending:
+            self.approving_date = timezone.now()
+            self.status = StatusOptions.Approved
+
+        self.save()
+
+    def cancel_order(self):
+        print(self.status)
```
ormergi, line 74: Is it a leftover?

orders/models.py
ormergi: Dont forget to follow up :slightly_smiling_face:

orders/tests.py
```diff
@@ -0,0 +1,61 @@
...
+        try:
+            Order.objects.get(id=-1)
+            assert False
+        except ObjectDoesNotExist:
+            assert True
+
+    def test_approve_order(self, order):
+
+        order.approve_order()
+        order.refresh_from_db()
```
ormergi, line 50: Why is it necessary to call refresh_from_db()?

orders/models.py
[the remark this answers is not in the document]
ormergi: Note that this should be cover by tests.

orders/tests.py
```diff
@@ -0,0 +1,61 @@
...
+        assert order in Order.objects.all()
+        order.delete()
+        assert order not in Order.objects.all()
+
+    def test_get_not_existed_order_by_order_id(self):
+        try:
+            Order.objects.get(id=-1)
+            assert False
+        except ObjectDoesNotExist:
+            assert True
```
ormergi, line 45: This test is not mandatory, it actually tests Django 'get' internal mechanism.
We take for granted that using a model .object.get() will work, its basic functionally of the framework
if it doesnt work the framework is broken.

ormergi (changes requested):
Hi Ofir, 
Please fix the commit message body, it should include at least what described by the PR description.

This PR describes that it will be possible to see the orders history, by it seems that once a daycare **or** a dogowner
user is deleted the order is deleted as well, meaning the history is lost :slightly_smiling_face: 
Are you going to address it in this PR?

orders/models.py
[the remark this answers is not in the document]
Omeramsc: Check out `set` and `set_null` options.

orders/models.py
```diff
@@ -0,0 +1,79 @@
...
+    Canceled = 'C', 'Canceled'
+    Finished = 'F', 'Finished'
+    OnGoing = 'O', 'On Going'
+
+
+class Order(models.Model):
+    dog_owner_id = models.ForeignKey(DogOwner, on_delete=models.CASCADE)
+    daycare_id = models.ForeignKey(DayCare, on_delete=models.CASCADE)
+    booking_date = models.DateField(auto_now=True)
+    approving_date = models.DateField(blank=True, null=True)
```
Omeramsc, line 19: nit: `date_approved` or `approval_date` or `approved_on` are better naming conventions then using present progressive.

orders/models.py
```diff
@@ -0,0 +1,76 @@
...
+class Order(models.Model):
+    order_id = models.AutoField(primary_key=True)
+    dog_owner_id = models.ForeignKey(DogOwner, on_delete=models.CASCADE)
+    daycare_id = models.ForeignKey(DayCare, on_delete=models.CASCADE)
+    booking_date = models.DateField(auto_now=True)
+    approving_date = models.DateField(blank=True, null=True)
+    starting_date = models.DateField()
+    finishing_date = models.DateField()
+    canceling_date = models.DateField(blank=True, null=True)
+    total_price = models.IntegerField()
```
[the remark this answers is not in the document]
Omeramsc: I agree with Ofir, BUT- you can replace `total_price` with `price_per_day`, therefore making it atomic and keeping it flexible (what if they want to add another they with the same price?)

orders/models.py
```diff
@@ -0,0 +1,79 @@
...
+    status = models.CharField(max_length=1,
+                              choices=StatusOptions.choices,
+                              default='P'
+                              )
+
+    def __str__(self):
+        return f"Dog owner ID: {self.dog_owner_id}, daycare ID: {self.daycare_id}, status: {self.status}"
+
+    @staticmethod
```
Omeramsc, line 32: should be @classmethod

orders/models.py
Omeramsc: I think you should keep it as it is for the time being to avoid complications, but do open an issue to research further on it later.

OfirMatasas: I've changed to Erez's advice. There are 2 methods in order for updating order's status - approve_order and cancel_order

orders/models.py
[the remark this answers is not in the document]
OfirMatasas: I think that for keeping it MVP, we should not support a situation when we keep the order when one of the users gets deleted.
Also, IMO, the idea of setting the foreign key to null has no logic in it.
If we do accept null as a proper value for the dog owner / daycare, we're saying that it's ok for the order to not have 2 sides of "the deal".

orders/models.py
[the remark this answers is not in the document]
OfirMatasas: We've talked about this scenario, and decided that editing the order is not optional.
If the dog owner wants to add a few more days to the order, he needs to make a new reservation for the extra days.
In a different scenario, if the daycare is not satisficed with the order before approving it, he needs to cancel it and make the dog owner create a new one.

orders/models.py
OfirMatasas: Yes is it. Removing it.

orders/tests.py
OfirMatasas: Got it. Removing it.

orders/models.py
OfirMatasas: Changed it to approval_date and cancellation_date.

orders/models.py
OfirMatasas: Doesn't a class method mean that we need to have an Order instance before calling the "create" method?
I've made it static, so the creation of an order is not dependent on a previous order instance.

orders/tests.py
OfirMatasas: Because the order instance before 

> Hi Ofir, Please fix the commit message body, it should include at least what described by the PR description.
> 
> This PR describes that it will be possible to see the orders history, by it seems that once a daycare **or** a dogowner user is deleted the order is deleted as well, meaning the history is lost 🙂 Are you going to address it in this PR?

I'll change the commit, but I've mentioned my opinion about deletion of a user and its affect on orders in another conversation here :)

orders/models.py
Omeramsc: No, a class method is a method of the class, and not of a specific object instance.

orders/tests.py
OfirMatasas: > Why is it necessary to call refresh_from_db()?

It not. I've used it to check if it's the cause of a failing test, but it wasn't.
Removed.

orders/models.py
OfirMatasas: Changed it.

orders/models.py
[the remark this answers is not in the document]
Omeramsc: You do not accept null usually, but only in cases like this when one user gets deleted. then, the view can show an "unavailable user" whenever it is null.

You can also create a default "unavailable user" to set it as, but I think this is not the right approach.

orders/models.py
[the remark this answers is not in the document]
Omeramsc: That's acceptable, but do note you won't have access to the price-per-day during the time of the order. (you can eventually calculate it, but it should be done the other way around).

orders/models.py
Omeramsc: huh, ok, I thought Erez suggested something else entierly. I agree with the way it is right now.

ErezCohenn (approved):

tamirmatok (approved):
Great job ! @OfirMatasas

ormergi (changes requested):

ormergi:
@OfirMatasas what about https://github.com/redhat-beyond/Hotails/pull/46#discussion_r843708882 
are you going to change them?

OfirMatasas:
> 



> @OfirMatasas what about [#46 (comment)](https://github.com/redhat-beyond/Hotails/pull/46#discussion_r843708882) are you going to change the other ones?

I've changed the name of "approving_date" to "approval_date" and "canceling_date" to "cancellation_date", as requested.

orders/tests.py
```diff
@@ -0,0 +1,62 @@
...
+    def test_cancel_order(self, test_order):
+        test_order.cancel_order()
+
+        assert test_order.status == StatusOptions.Canceled
+        assert test_order.cancellation_date is not None
+
+    def test_total_price_of_order(self, test_order):
+        assert test_order.get_order_total_price() == PRICE_PER_DAY * NUMBER_OF_DAYS
+
+    def test_dog_owner_foreign_key_after_deletion_of_dog_owner(self, test_order):
```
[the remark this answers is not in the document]
OfirMatasas: > I would call it `test_dog_owner_id_is_deleted_when_dog_owner_is_deleted` (same for daycare user case)

Changed it.

orders/tests.py
```diff
@@ -0,0 +1,62 @@
...
+
+NUMBER_OF_DAYS = 3
+STARTING_DATE = timezone.now()
+FINISHING_DATE = timezone.now() + datetime.timedelta(days=NUMBER_OF_DAYS)
+PRICE_PER_DAY = 100
+STATUS = StatusOptions.Pending
+
+
+@pytest.fixture
+def test_order():
```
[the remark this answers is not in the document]
OfirMatasas: > Note that this is not a test as nothing is being checked, it is a fixture for creating an on order please change its name accordingly.

Changed it according to previous pull requests to "create_order".

ormergi:
> I've changed the name of "approving_date" to "approval_date" and "canceling_date" to "cancellation_date", as requested.

@OfirMatasas  thanks for the changes.
Please change `booking_date`, starting_date` and `finishing_date` to `book_date`, `start_date` and `end_date` as well, 
it is the common name convention for a class attributes.

orders/tests.py
```diff
@@ -0,0 +1,62 @@
...
+                        starting_date=STARTING_DATE,
+                        finishing_date=FINISHING_DATE,
+                        price_per_day=PRICE_PER_DAY,
+                        )
+
+
+@pytest.mark.django_db
+class TestOrderModel:
+    def test_create_order(self):
+        new_order = Order.create(dog_owner_id=DogOwner.objects.get(id=1),
```
Omeramsc, line 29: why do we create a new order here instead of using the fixture above?

orders/tests.py
```diff
@@ -0,0 +1,62 @@
...
+        assert create_order.status == StatusOptions.Approved
+        assert create_order.approval_date is not None
+
+    def test_cancel_order(self, create_order):
+        create_order.cancel_order()
+
+        assert create_order.status == StatusOptions.Canceled
+        assert create_order.cancellation_date is not None
+
+    def test_total_price_of_order(self, create_order):
```
Omeramsc, line 53: a better implementation would include a parametrization with multiple combinations of PRICE_PER_DAY and NUMBER_OF_DAYS, checking we get what we expected even though we tried to fail the function

orders/tests.py
```diff
@@ -0,0 +1,62 @@
+import datetime
+from django.utils import timezone
+from dogowner.models import DogOwner
+from daycare.models import DayCare
+from .models import Order
+from .models import StatusOptions
+import pytest
+
+NUMBER_OF_DAYS = 3
```
Omeramsc, line 9: since those should be used only on `create_order`, I see no need for them to be constants

Omeramsc (changes requested):
Left some minor comments, let's fix them today and get this merged. feel free to ask if something is unclear

orders/tests.py
OfirMatasas: > since those should be used only on `create_order`, I see no need for them to be constants

There's a test that calculate the total price of the order, and it's using this constant as well

orders/tests.py
OfirMatasas: You're right, I'll change the test to check if the creation of the fixture succeeded.

orders/tests.py
OfirMatasas: > a better implementation would include a parametrization with multiple combinations of PRICE_PER_DAY and NUMBER_OF_DAYS, checking we get what we expected even though we tried to fail the function

Could you give an example for better implementation of the test? 
I don't think I understand what you mean by "multiple combinations of PRICE_PER_DAY and NUMBER_OF_DAYS", if it mean to check the same order with right and wrong parameters, or checking several orders.

orders/tests.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def test_cancel_order(self, create_order):
+        create_order.cancel_order()
+
+        assert create_order.status == StatusOptions.Canceled
+        assert create_order.cancellation_date is not None
+
+    @pytest.mark.parametrize('starting_date, finishing_date, price_per_day, num_of_days, expected_result',
+                             [
+                                 (timezone.now(), timezone.now() + datetime.timedelta(days=3), 100, 3, 300),
```
Omeramsc, line 49: You don't need to include the time function calls in the parametrization. use them under the `create` call. the only thing that changes from test to test (in the time aspect) is the number of days. you only need to keep this parameter.

orders/tests.py
```diff
@@ -0,0 +1,70 @@
...
+from dogowner.models import DogOwner
+from daycare.models import DayCare
+from .models import Order
+from .models import StatusOptions
+import pytest
+
+STARTING_DATE = timezone.now()
+FINISHING_DATE = timezone.now() + datetime.timedelta(days=3)
+PRICE_PER_DAY = 100
+STATUS = StatusOptions.Pending
```
Omeramsc, line 12: None of those constants is used more than once, you can just use them in the dedicated place instead of declaring them here (for now I don't see a benefit for it)

OfirMatasas: > None of those constants is used more than once, you can just use them in the dedicated place instead of declaring them here (for now I don't see a benefit for it)

Changed it.

orders/tests.py
OfirMatasas: > You don't need to include the time function calls in the parametrization. use them under the `create` call. the only thing that changes from test to test (in the time aspect) is the number of days. you only need to keep this parameter.

Removed both starting_date and finishing_date from the parametrization, and used the num_of_days for delta.

Omeramsc (approved):
Awesome work, LGTM!

OfirMatasas:
> > I've changed the name of "approving_date" to "approval_date" and "canceling_date" to "cancellation_date", as requested.
> 
> @OfirMatasas thanks for the changes. Please change `booking_date`, starting_date`and`finishing_date`to`book_date`, `start_date`and`end_date` as well, it is the common name convention for a class attributes.

Changed, according to your request :)

ormergi (approved):
Thanks for the changes, it looks good!

PR #56, opened 2022-04-06
Creating a navigation bar for Hotails templates

### LINKED ISSUE #55: Create a base template for Hotails
Create a base template for Hotails, containing a navigation bar (navbar), so all templates will have the same one.
The navbar will contain:

- Hotails logo
- Profile
- Chat
- Orders
- About
- Logout option

OfirMatasas:
This HTML file will be the base of all other templates on Hotails.
Which means that every HTML page we'll create in Hotails, will expend
this HTML, so they all will have the same navigation bar.

A picture of the navbar:
Untitled

closes #55 


tamirmatok (approved):
Great work ! @OfirMatasas

ErezCohenn (changes requested):
Nive job @OfirMatasas 
few things:
- there is a missing `Search` field as we designed [here](https://excalidraw.com/#room=725b3d8abe435444c7c7,yk6cMq1PB8hPNThZjRyiiA) in the navbar
- I think you should transfer the exists content home-page to the about page, because we are going to run over the home page with daycare-home and dowonwer-home.
you should instead put in the body the content `Coming soon` or something like that

OfirMatasas:
> Nice job @OfirMatasas few things:
> 
> * there is a missing `Search` field as we designed [here](https://excalidraw.com/#room=725b3d8abe435444c7c7,yk6cMq1PB8hPNThZjRyiiA) in the navbar
> * I think you should transfer the existing content home page to the about page because we are going to run over the home page with daycare-home and dowonwer-home.
>   you should instead put in the body the content `Coming soon` or something like that
> * suggestion: if we change the UI in the project, maybe we need to add some screenshots in the PR description that will make it easier for the reviewers

1. I thought we decided to remove the search option from the navbar, since the daycare don't have anything to search for on Hotails, and keep the search in dog owner's homepage, along the relevant daycares. I can add it.
2. Good idea. I'll change that.
3. I think the pictures should attach to the description of the PR, and not the PR itself, like I just did here, because it's not a part of the project. I think we should upload only the changes in the UI, like before and after the PR.

ErezCohenn:
> > Nice job @OfirMatasas few things:
> > 
> > * there is a missing `Search` field as we designed [here](https://excalidraw.com/#room=725b3d8abe435444c7c7,yk6cMq1PB8hPNThZjRyiiA) in the navbar
> > * I think you should transfer the existing content home page to the about page because we are going to run over the home page with daycare-home and dowonwer-home.
> >   you should instead put in the body the content `Coming soon` or something like that
> > * suggestion: if we change the UI in the project, maybe we need to add some screenshots in the PR description that will make it easier for the reviewers
> 
> 1. I thought we decided to remove the search option from the navbar, since the daycare don't have anything to search for on Hotails, and keep the search in dog owner's homepage, along the relevant daycares. I can add it, but the daycare won't use it.
> 2. Good idea. I'll change that.
> 3. I think the pictures should attach to the description of the PR, and not the PR itself, like I just did here, because it's not a part of the project. I think we should upload only the changes in the UI, like before and after the PR.

1. as we talked offline, we decided to add to the navbar the `search` field

ErezCohenn (changes requested):
Don't forget to add tests :)

Hotails/urls.py
```diff
@@ -19,5 +19,6 @@
 
 urlpatterns = [
     path('admin/', admin.site.urls),
-    path('', views.homepage, name='homepage')
+    path('', views.homepage, name='homepage'),
+    path('about/', views.about, name='about'),
```
jeniaSakirko, line 23: I think best will be to add all pages to the urls and direct all to the homepage view

main/templates/main/homepage.html
```diff
@@ -1,7 +1,6 @@
-<!DOCTYPE html>
+{% extends "main/base_template.html" %}
+{% block content %}
 {% load static %}
-<html style="font-size: 16px;">
-
 <head>
```
Omeramsc, line 4: Do we need to keep the head and body tags?

Omeramsc (changes requested):

main/templates/main/homepage.html
OfirMatasas: I've removed the head and body and put it all into "container" for keeping using the class and other stylings.

main/templates/main/homepage.html
```diff
@@ -1,32 +1,15 @@
...
-  <meta property="og:title" content="Home">
-  <meta property="og:type" content="website">
-</head>
-
-<body data-home-page="index.html" data-home-page-title="Home" class="u-body u-xl-mode">
-  <header class="u-clearfix u-header u-header" id="sec-23af"><a href="" class="u-image u-logo u-image-1"
-      data-image-width="300" data-image-height="300">
-      <img src="{% static 'images/Hotails-logo.jpg' %}" class="u-logo-image u-logo-image-1">
-    </a></header>
+<container data-home-page="index.html" data-home-page-title="Home" class="u-body u-xl-mode">
```
Omeramsc, line 12: never heard of a container tag. is that a thing? I know that in order to contain content in a tag with no special affiliation you'd usually use a <div> (while container is a class of bootstrap and not a tag)

OfirMatasas: > never heard of a container tag. is that a thing? I know that in order to contain content in a tag with no special affiliation you'd usually use a div (while container is a class of bootstrap and not a tag)

I guess that "div" would work, but "container" is the same thing for having a "box of information" under the same "roof", for having the option to apply styling easier than specify the "div" you're meaning to.

Yuval-Vino (approved):
LGTM!

main/templates/main/homepage.html
Omeramsc: I see, good to know. :)

Omeramsc (approved):
LGTM good job

PR #64, opened 2022-04-08
Adding Homepage for daycare user

ErezCohenn:
Signed-off-by: Erez Cohen <erez6356458@gmail.com>

### Desired Outcome

After the daycare user is logged in to the website, he moved to the homepage of the site.
The homepage is the first page that the user exposed after login and described the main details for the daycare such as the details of the daycare and the orders that are waiting for approval

### Implemented Changes

Creating homepage **only** for daycare users, if another types of users will try to reach this site the will redirect back to the login page.

### Connected Issue/Story

Resolves #63 
Depends #56 , #62 , #75, #84

### Definition of Done

- [x] arrange the URL patterns in Hotails and main apps
- [ ] develop logic functionality for daycare users
- [ ] Adding a UI for visualization to logical implementations
- [ ] testing

#### Test coverage

- [x] This PR includes new unit and integration tests to go with the code
  changes
- [ ] The changes in this PR do not require tests

#### Documentation

- [ ] Docs (e.g. `README`s) were updated in this PR
- [ ] A follow-up issue to update official docs has been filed here: [insert issue ID]()
- [x] This PR does not require updating any documentation

PR #67, opened 2022-04-10
Adding homepage for dog owner

### LINKED ISSUE #66: Create homepage for dog owner
## Current Behavior
Currently, the dog owner does not have a customized homepage.

## Expected Behavior
When the dog owner is logged in to Hotails, or when the home is pressed via navbar, All the daycares in the database will be shown to the dog owner on its homepage, with the possibility to check their profile.

### Steps to Reproduce

1. Extend base_template for navbar
2. Pull all daycares from database
3. Show all daycares in the dog owner
4. Show dog owner profile on the sidebar

OfirMatasas:
## Desired Outcome
After the dog owner user is logged in to the website, he moved to the homepage of the site.
The homepage is the first page that the user exposed after login (or after clicking the Hotails logo on the navbar), where the dog owner can see all the daycares registered to Hotails, and check their profile for more actions.

## Implemented Changes
Creating homepage only for dog owners, if another types of users will try to reach this site the will redirect back to the login page.

## Connected Issue/Story
closes #66 

## Definition of Done
 Arrange the URL patterns in Hotails and main apps
 Develop logic functionality for dog owner users
 Add a UI for visualization to logical implementations testing

## Test coverage
 This PR includes new unit and integration tests to go with the code changes

Untitled


ormergi:
ping

dogowner/templates/dogowner/dog_owner_homepage.html
```diff
@@ -0,0 +1,26 @@
...
+{% block stylesheets %}
+    <link rel="stylesheet" href="{% static 'CSS/dog_owner_homepage.css' %}">
+{% endblock %}
+
+{% block content %}
+<div class="row row-cols-1 g-3 cards">
+  <div style="display: flex; flex-direction: row; flex-wrap: wrap">
+    {% for daycare in daycares %}
+    <div class="card">
+      <img src="{{ daycare.get_daycare_primary_image }}" alt="{{ daycare.name }} image" class="card-img-top">
```
ErezCohenn, line 13: In the daycare model, you added the method: `get_daycare_primary_image` that could return null, 
so in that case, you are not referring to that case in the html file.
optional solution: add a field 'daycare_primary_image' in 'dogowner/views.py' in the context array, and there check if is a null, or check it in the html file

ErezCohenn (commented):
Great Job tamir and ofir!!
Just one thing for next: I think in the future when we will implement the search app, the dog owner will see on the home page the most rated daycares, instead of all the daycares that exsists.

OfirMatasas:
> Great Job tamir and ofir!! Just one thing for next: I think in the future when we will implement the search app, the dog owner will see on the home page the most rated daycares, instead of all the daycares that exsists.

You're right, but the total review rating of a daycare does not calculated in Hotails ATM.
When we'll implement this method, we'll change the order of presented daycares to the dog owner on his homepage and/or on searching.

dogowner/templates/dogowner/dog_owner_homepage.html
OfirMatasas: You're right, but the total review rating of a daycare does not calculated in Hotails ATM.
When we'll implement this method, we'll change the order of presented daycares to the dog owner on his homepage and/or on searching.

jeniaSakirko (approved):
LGTM

dogowner/views.py
```diff
@@ -1,3 +1,10 @@
...
+from daycare.models import DayCare
 
-# Create your views here.
+
+def dog_owner_home(request):
+    context = {
+        'dogowner': request.user,
+        'daycares': DayCare.objects.all(),
+    }
+    return render(request, 'dogowner/dog_owner_homepage.html', context)
```
tamirmatok, line 10: Im not sure we need to send 'dogowner' containg User in context because it already include in the request .
In addition , we should use @login_requierd decorater for case when User try to reach this page from URL .

dogowner/templates/dogowner/dog_owner_homepage.html
Omeramsc: You must deal with the case you get a None as the image. You can place a placeholder "no image available" or something, but not a broken link.

OfirMatasas: I've added on #77 2 default profile pictures, 1 for daycare and 1 for dog owner.
I'll add it on this PR as well and I'll ping you.

static/CSS/dog_owner_homepage.css
```diff
@@ -0,0 +1,46 @@
...
+
+.card-img-top
+{
+    height: 200px;
+}
+
+.btn
+{
+
+}
```
Omeramsc, line 46: Is that necessary?

daycare/models.py
```diff
@@ -50,6 +50,11 @@ def create(email, username, password, name, description, price_per_day, capacity
 
         return new_daycare
 
+    def get_daycare_primary_image(self):
```
Omeramsc, line 53: I don't see a test for this

static/CSS/dog_owner_homepage.css
OfirMatasas: it's not. Deleted.

dogowner/views.py
OfirMatasas: Done.

daycare/test_image.py
```diff
@@ -20,3 +22,13 @@ def test_image_creation_with_invalid_image_url(self, create_daycare_user):
...
+
+    def test_daycare_has_customized_profile_image(self, create_image1, create_image2, create_daycare_user):
+        daycare_profile_image = create_daycare_user.get_daycare_primary_image_url()
+        assert daycare_profile_image != DEFAULT_DAYCARE_PROFILE_URL
+        assert daycare_profile_image is not None
+
+    def test_daycare_has_default_profile_image(self, create_daycare_user):
+        daycare_profile_image = create_daycare_user.get_daycare_primary_image_url()
+        assert daycare_profile_image == DEFAULT_DAYCARE_PROFILE_URL
+        assert daycare_profile_image is not None
```
Omeramsc, line 34: The second assertion is redundant. we just confirmed the profile is the default, so of course, it is not None.

daycare/test_image.py
```diff
@@ -20,3 +22,13 @@ def test_image_creation_with_invalid_image_url(self, create_daycare_user):
         with pytest.raises(ValidationError,
                            match="Invalid URL image - URL should end with \'.gif\', \'.png\', \'.jpg\' or \'.jpeg\'."):
             Image.create(url="NOT_VALID_URL", daycare_id=DayCare.objects.get(id=create_daycare_user.id))
+
+    def test_daycare_has_customized_profile_image(self, create_image1, create_image2, create_daycare_user):
+        daycare_profile_image = create_daycare_user.get_daycare_primary_image_url()
+        assert daycare_profile_image != DEFAULT_DAYCARE_PROFILE_URL
+        assert daycare_profile_image is not None
+
+    def test_daycare_has_default_profile_image(self, create_daycare_user):
```
Omeramsc, line 31: would expect a better name. *why* does it have the default profile_image? (test_......_when_______)

daycare/models.py
```diff
@@ -50,6 +49,12 @@ def create(email, username, password, name, description, price_per_day, capacity
 
         return new_daycare
 
+    def get_daycare_primary_image_url(self):
```
Omeramsc, line 52: maybe there should be a better method for a daycare user to decide when is his primary photo. 
that can wait, but worth thinking about.

main/views.py
```diff
@@ -13,7 +15,8 @@ def index(request):
 
 @login_required()
 def homepage(request):
-    return render(request, 'main/homepage.html')
+    if DogOwner.objects.filter(user=request.user).exists():
```
Omeramsc, line 18: And what happens if it is not a dog_owner?
anyway, this needs a test.

ErezCohenn (approved):
LGTM !

daycare/test_image.py
OfirMatasas: Changed.

daycare/models.py
OfirMatasas: You're right. I thought about it, but for now, as we want to speed things up, this current state is good enough, and (hopefully) will be changed.

main/views.py
Omeramsc: My question was aimed toward the fact that *right now* the homepage leads somewhere only if the user is a DogOwner.
if there is only a dogowner, no real need for this statement.
if there is not only a dogowner, this needs an "elif". 
OR... maybe it should lead to a dedicated landing\about page.

daycare/models.py
Omeramsc: We don't "hope", we do. :) please open an issue for it and tag it here.

daycare/test_image.py
Omeramsc: great!

daycare/models.py
OfirMatasas: Created #93 :)

main/views.py
OfirMatasas: I've added an elif statement for the daycare, redirected (for now) to the "Coming soon" homepage,
and if the logged in user is not a dog owner and not a daycare, they'll be redirected to the "about" page.

Omeramsc: There are 2 flaws here: 
1. If this stays as a `login_required`, there is no need for the last return. do you want a not-logged-in user to not be able to view the homepage?
2. missing a test.

tamirmatok: 1) Fixed . 
2) Test for both cases were added.

main/tests.py
```diff
@@ -82,8 +81,20 @@ def test_root_entrypoint_redirection_logged_user(self, client, create_dog_owner_
...
     def test_unlogged_user_access_to_homepage(self, client):
         response = client.get("/homepage/")
         assert response.status_code == 302
         assert response['Location'] == '/login/?next=/homepage/'
+
+    def test_dog_owner_homepage_is_visible_for_dog_owner(self, client, create_dog_owner_user):
+        client.force_login(user=create_dog_owner_user.user)
+        response = client.get("/homepage/")
+        assert response.status_code == 200
+        assert response.context['daycares']
```
Yuval-Vino, line 95: what are we testing here?

main/tests.py
```diff
@@ -82,8 +81,20 @@ def test_root_entrypoint_redirection_logged_user(self, client, create_dog_owner_
...
+
+    def test_day_care_homepage_is_visible_for_day_care(self, client, create_daycare_user):
+        client.force_login(user=create_daycare_user.user)
+        response = client.get("/homepage/")
+        assert response.status_code == 200
```
Yuval-Vino, line 100: is this test relevant for this PR? seems like #64 should take care of it, as it does

Yuval-Vino (commented):
Good job guys! i left few little comments tell me what you think!

main/tests.py
tamirmatok: We are testing successful enter to dog_owner_home page and that the response containing 'daycares' sent from dog_owner_home view .

main/tests.py
tamirmatok: If I understand it right , according to https://github.com/redhat-beyond/Hotails/pull/67#discussion_r862435378 this test is requierd for now.

main/tests.py
Yuval-Vino: If I understand correctly,  here we only test that the HTML has daycares argument, isn't it redundant?
because this argument surely gets passed from views.py. 
Maybe it's better to check the value (the logic in views.py) of this argument is really what we expect

tamirmatok: 1) Im not sure it is redundant because if the user is day_care he will not have daycares in response.context -
so in that way I can make sure that the current URL is in dog_howner_home.
2)Can you share what r u thinking about ? the logic in the home_page_view for now is to send all daycares with the request.

main/tests.py
Omeramsc: IMO Yuval is right, this does not seem related to this PR (we rate dealing with dog owner homepage, not day care)

main/tests.py
Omeramsc: I agree, we should not only check the existence of "daycares" in the context, but the daycares information in the context as well.

OfirMatasas: Added a check for the context of the "daycares" from the response.

main/tests.py
OfirMatasas: Removed this test.

main/tests.py
```diff
@@ -64,8 +65,15 @@ def test_root_entrypoint_redirection_logged_daycare_user(self, client, create_da
...
     def test_unlogged_user_access_to_homepage(self, client):
         response = client.get("/homepage/")
         assert response.status_code == 302
         assert response['Location'] == '/login/?next=/homepage/'
+
+    def test_dog_owner_homepage_is_visible_for_dog_owner(self, client, create_dog_owner_user):
+        client.force_login(user=create_dog_owner_user.user)
+        response = client.get("/homepage/")
+        assert response.status_code == 200
+        assert list(response.context['daycares']) == list(DayCare.objects.all())
```
Omeramsc, line 79: I want this PR merged ASAP, so please open a ticket to: assign both sides of the assertion equation to variables, then assert between the variables. it will improve the readability of the test.

Omeramsc (approved):
LGTM

main/tests.py
OfirMatasas: Opened #104

PR #60, opened 2022-04-06
🖼️ Adding the possibility for a daycare owner to upload images to his daycare

ErezCohenn:
Signed-off-by: Erez Cohen <erez6356458@gmail.com>

### Desired Outcome

Updating the daycare model by adding the class`image` that will enable the daycare owner to upload images to his daycare.

### Implemented Changes

Updating the daycare model by adding the field `image` 

### Connected Issue/Story

Resolves #59 

### Definition of Done
- [x] Adding Image class to handle multiply images per daycare (by using FK for daycare id)
- [x] update the daycare migrations
- [x]  testing the new model Image
- [x]  update the ERD diagram

#### Changelog

- [ ] The CHANGELOG has been updated
- [x] This PR does not include user-facing changes and doesn't require a
  CHANGELOG update

#### Test coverage

- [x] This PR includes new unit and integration tests to go with the code
  changes
- [ ] The changes in this PR do not require tests

#### Documentation

- [ ] Docs (e.g. `README`s) were updated in this PR
- [ ] A follow-up issue to update official docs has been filed here: [insert issue ID]()
- [x] This PR does not require updating any documentation

PR #65, opened 2022-04-10
Add static dog-owner users to dogowner app.

tamirmatok:
### Desired Outcome
Add static dog-owner user information to fulfill Hotails app with relevant information.

### Implemented Changes
- Modify dog_picture_url field form CharField to UrlField .
- Add image_url validator with test .
- Add static dog-owner migration file.

### Connected Issue/Story
closes #58

#### Test coverage

- [x] This PR includes new unit and integration tests to go with the code
  changes
- [ ] The changes in this PR do not require tests

#### Documentation

- [ ] Docs (e.g. `README`s) were updated in this PR
- [ ] A follow-up issue to update official docs has been filed here: [insert issue ID]()
- [x] This PR does not require updating any documentation

PR #69, opened 2022-04-11
👤 Add Static daycares users for daycare app

### LINKED ISSUE #68: Add static Daycare users
## Expected Behavior

Because the project uses only static users without the ability to register at the moment, there is a need for a large number of static users of daycare owners for regular work.

## Current Behavior

There are only 3 static daycare users in the data migrations

### Steps to Reproduce

- [ ] Create static Daycare users.
- [ ] Create static images for the daycares users
- [ ] Modify data migrations in daycare app
- [ ] Add resource folder with daycare information

ErezCohenn:
### Desired Outcome

Because the project uses only static users without the ability to register at the moment, there is a need for a large number of static users of daycare owners for regular work.

### Implemented Changes

creating static daycare users and static images for the daycares that implemented in the data migrations in daycare app

### Connected Issue/Story

Resolves #68 
Depends #60 

### Definition of Done

- [x] Create static Daycare users.
- [x] Create static images for the daycares users
- [x] Modify data migrations in daycare app
- [x] Add resource folder with daycare information

#### Changelog

  CHANGELOG update

#### Test coverage

  changes (the existing tests in the daycare app are testing the new data)
- [] The changes in this PR do not require tests

#### Documentation


daycare/resources/daycare_static_users.py
```diff
@@ -0,0 +1,65 @@
+USERNAME = 'staticUser'
```
Omeramsc, line 1: as I mentioned in #65 , I don't really like this solution, as it'll be better to mention those explicitly and in one place, instead of following loops through two different files.

ErezCohenn: The file `daycare_static_users` with the static information is very cluttered, so if all this information sat in only one place as you mentioned, would it not be in this case that it creates more clutter, and is less readable and understandable?
I would love to know what you mean by saying that it is best to have everything in one place, do you mean that it should be in the data migration file (that will include the all static data), or in the `daycare_static_users.py` file the will include all the logic and the loops that are in the data migration right now?

Omeramsc: Think about the (new) developer experience, trying to either create or understand a test using some of the daycares created from this information. 
since it is all created within loops, he'll have to match the different parameters according to their indexes to understand what he should expect. e.g: which email goes with which username, pass, name, etc...
this is not readable.
I much prefer having a longer and more cluttered file with a list of each daycare object parameters, so I'll understand in one look what I see and which daycare has which property.

daycare/migrations/0005_image_data_migration.py
```diff
@@ -0,0 +1,28 @@
...
+    ]
+
+    def generate_data(apps, schema_editor):
+        from daycare.models import DayCare, Image
+
+        test_data = []
+
+        for daycare in DayCare.objects.all():
+            for i in range(1, 5):
+                test_data.append((random.choice(DAYCARE_IMAGE_LIST), daycare))
```
Omeramsc, line 20: I would avoid using anything random in test data... it should be explicit.

daycare/resources/daycare_static_users.py
ErezCohenn: And what about if I will transfer all the static data from `resources/daycare_static_users.py` to the data migration,
and there I will still use the loop, that each static daycare user, will get his data from the lists, that each list is a field in the model.
For example (for the name field):
DAYCARE_NAME = [name1, name2, ...., ...nameN]
and I will run in a loop for N static users, and for each user: daycare.name = DAYCARE_NAME[i].
I think in that way it will be readable and easy to understand for the reader, instead of creating N lists, that each list will include all the fields of the user (that will be a little messy).
@Omeramsc  tell me what u think

OfirMatasas (approved):
Looks good overall.
The only change I'll apply is to remove the constants as Omer suggested, and keep only the URLs of the images for the tests.

ErezCohenn:
> Looks good overall. The only change I'll apply is to remove the constants as Omer suggested, and keep only the URLs of the images for the tests.

notice that this PR is dependent on #60, so the comment is  not relevant here :)

daycare/resources/daycare_static_users.py
ErezCohenn: Got your point.
So I created the static daycare users only in the data migrations,
but what do you think about the static images data? It does not really matter what pictures each static user will get so I thought it would be right to leave the situation as it is, I would love to know what you think about it.

Omeramsc: It doesn't matter now, but what about the future?
anyway, don't use any "random" element in tests. even if it doesn't matter. tests has to be reliable, with specific data expected in and out.

ErezCohenn: I don't think it will matter in the future also, because this data is static, and has no effect on future "real" users that will log in to the website.
about the random images, i will change it.

daycare/migrations/0005_image_data_migration.py
ErezCohenn: @Omeramsc you can take another look, I changed it that the static images will be more explicit

daycare/migrations/0005_image_data_migration.py
```diff
@@ -0,0 +1,108 @@
...
+        ]
+
+        static_daycare_id = []
+
+        for daycare in DayCare.objects.all():
+            if 'static' in daycare.user.username:
+                static_daycare_id.append(daycare)
+
+        static_daycare_images = [
+            (static_daycare_id[0], DAYCARE_IMAGE_LIST[0]),
```
Omeramsc, line 40: In this *specific* case, if we're going with this implementation, I'd suggest using a loop with enumurate instead

daycare/migrations/0005_image_data_migration.py
```diff
@@ -0,0 +1,50 @@
...
+            "https://www.flyingfurintheburg.com/wp-content/uploads/small-dog-daycare-4.jpg",
+            "https://www.flyingfurintheburg.com/wp-content/uploads/small-dog-daycare-5.jpg"
+        ]
+
+        static_daycare_id = []
+        static_daycare_images = []
+
+        for daycare in DayCare.objects.all():
+            if 'static' in daycare.user.username:
+                static_daycare_id.append(daycare)
```
Omeramsc, line 38: You can use [list comprehension](https://www.w3schools.com/python/python_lists_comprehension.asp) instead, making it more pythonic, instead of declaring an empty list first, c-style.

daycare/migrations/0005_image_data_migration.py
```diff
@@ -0,0 +1,50 @@
...
+
+        for index, daycare in enumerate(static_daycare_id):
+            for i in range(5):
+                static_daycare_images.append((daycare, DAYCARE_IMAGE_LIST[(index + i) % len(DAYCARE_IMAGE_LIST)]))
```
Omeramsc, line 42: Sorry, but I don't understand what is happening here or why (why are we calculating stuff?)

ErezCohenn: instead of random, every static daycare user will get 5 images, so that each image will be selected by the daycare index combined with the image index, and in order not to exceed the size of the image set, I added a modulo.
so that daycare01 will get images 1,2,...,5, daycare2 will get images 2,3...,6, and so on.
actually, I implemented the explicit data in the previous push, to run in a loop as you commented.

do you recommend any other effective solution instead of random, this way, or the last method I used with explicit?

Omeramsc: At this point, things are over-complicated. the fact that we need to think over a trivial addition like this for so long means there is a design flaw. 
If we're set on this "static daycare" implementation, I'd explicitly have photo X attached to daycare Y. whether it means having the same photo for multiple daycares, more photos, or less daycares, is up to you. but it needs to be clear and explicit enough so "my grandmother will be able to understand it".

ErezCohenn: I decided to make the things more simple because its only static data so it doesn't really matter which daycare will get each image, so every daycare will get all the images in the static images list by using
`list comprehension`.

Yuval-Vino (approved):
LGTM!

jeniaSakirko (approved):
LGTM

tamirmatok (approved):
LGTM

daycare/migrations/0005_image_data_migration.py
Omeramsc: Never say never. even though it is static data, it does matter. 
I'm okay with the current implementation, while having all the photos is less reliable than having specific ones mentioned, it is a good-enough compromise and it is explicit enough.

The list comprehension is also well done! 
I'm not sure the second one is absolutely needed, but it is good enough as is

good job!

PR #75, opened 2022-04-17
Search for review by daycare

### LINKED ISSUE #74: Search for review by daycare id
## Expected Behavior

There is an option to search for review by daycare id

## Current Behavior

The is no option to search for review by daycare id

## Acceptance Criteria

Tests were added to search for review by daycare id - the tests are not depending on existing info

jeniaSakirko:
### Desired Outcome

The review model will have option to search by daycare id

### Implemented Changes

review/migrations/0002_test_data.py - Fix the way review data was created to use the create method
review/models.py - Add option to get all reviews by daycare id
review/tests.py - Add tests for the new method 

### Connected Issue/Story

Resolves #74 

### Definition of Done

There is an option to get all reviews for specific daycare
Tests were added to search for review by daycare id - the tests are not depending on existing info

#### Changelog

  CHANGELOG update

#### Test coverage

  changes
- [ ] The changes in this PR do not require tests

#### Documentation

### DIFF
--- a/review/conftest.py
+++ b/review/conftest.py
@@ -1,0 +1,33 @@
+import pytest
+from daycare.models import DayCare
+from dogowner.models import DogOwner
+from review.models import Review
+
+
+@pytest.fixture
+def daycare():
+    return DayCare.create(email="daycare_review@gmail.com", username="daycare_review", password="pass",
+                          name="reviewDayCare", description="This is test daycare for review", price_per_day=10,
+                          capacity=50, area="Merkaz", city="Tel-Aviv", address="The best review")
+
+
+@pytest.fixture
+def dogowner():
+    return DogOwner.create(email="dogowner_review@gmail.com", username='dogowner_review', password='pass',
+                           dog_name='dogReview', first_name='firstNameReview', last_name='lastNameReview',
+                           phone_number=1234567890, dog_race='review',
+                           dog_picture_url='https://www.google.com/review.jpg', dog_age=1, dog_weight=1, dog_gender='M')
+
+
+@pytest.fixture
+def review_data():
+    pytest.REVIEW = 'sample review'
+    pytest.RATING = 5
+    pytest.DAY_CARE_ID = 1
+    pytest.DOG_OWNER_ID = 1
+
+
+@pytest.fixture
+def review(review_data):
+    return Review.create(review=pytest.REVIEW, rating=pytest.RATING, daycare_id=pytest.DAY_CARE_ID,
+                         dogowner_id=pytest.DOG_OWNER_ID)
--- a/review/tests.py
+++ b/review/tests.py
@@ -2,52 +2,48 @@ import pytest
 from .models import Review, DayCare, DogOwner
 from django.core.validators import ValidationError
 
-REVIEW = 'sample review'
-RATING = 5
-DAY_CARE_ID = 1
-DOG_OWNER_ID = 1
 
-
-@pytest.mark.django_db
-@pytest.fixture
-def test_review():
-    return Review.create(review=REVIEW, rating=RATING, daycare_id=DAY_CARE_ID, dogowner_id=DOG_OWNER_ID)
-
-
-@pytest.mark.django_db
+@pytest.mark.django_db()
 class TestReviewModel:
-    def test_persist_review(self, test_review):
-        assert test_review.review == REVIEW
-        assert test_review in Review.objects.all()
+    def test_persist_review(self, review, review_data):
+        assert review.review == pytest.REVIEW
+        assert review in Review.objects.all()
 
-    def test_del_review(self, test_review):
-        test_review.delete()
-        assert test_review not in Review.objects.all()
+    def test_del_review(self, review):
+        review.delete()
+        assert review not in Review.objects.all()
 
-    def test_rating_limit_min(self):
+    @pytest.mark.parametrize("rating", [-1, 6])
+    def test_rating_limit_min(self, rating, review_data):
         with pytest.raises(ValidationError):
-            Review.create(review=REVIEW, rating=-1, daycare_id=DAY_CARE_ID, dogowner_id=DOG_OWNER_ID)
+            Review.create(review=pytest.REVIEW, rating=rating, daycare_id=pytest.DAY_CARE_ID,
+                          dogowner_id=pytest.DOG_OWNER_ID)
 
-    def test_rating_limit_max(self):
-        with pytest.raises(ValidationError):
-            Review.create(review=REVIEW, rating=6, daycare_id=DAY_CARE_ID, dogowner_id=DOG_OWNER_ID)
-
-    def test_delete_related_daycare(self, test_review):
-        daycare = DayCare.objects.get(pk=DAY_CARE_ID)
+    def test_delete_related_daycare(self, review, review_data):
+        daycare = DayCare.objects.get(pk=pytest.DAY_CARE_ID)
 
         assert daycare in DayCare.objects.all()
-        assert test_review in Review.objects.all()
+        assert review in Review.objects.all()
 
         daycare.delete()
         assert daycare not in DayCare.objects.all()
-        assert test_review not in Review.objects.all()
+        assert review not in Review.objects.all()
 
-    def test_delete_related_dogowner(self, test_review):
-        dogowner = DogOwner.objects.get(id=DOG_OWNER_ID)
-
+    def test_delete_related_dogowner(self, dogowner, review, review_data):
         assert dogowner in DogOwner.objects.all()
-        assert test_review in Review.objects.all()
+        assert review in Review.objects.all()
 
         dogowner.delete()
         assert dogowner not in DogOwner.objects.all()
-        assert test_review in Review.objects.all()
+        assert review in Review.objects.all()
+
+    @pytest.mark.parametrize("number_of_reviews", [0, 1, 3])
+    def test_get_review_by_daycare_id(self, number_of_reviews, dogowner, daycare):
+        review_list = [Review.create(review='review', rating=1, daycare_id=daycare.id, dogowner_id=dogowner.id)
+                       for _ in range(number_of_reviews)]
+
+        reviews_by_id_list = Review.get_review_by_daycare_id(daycare.id)
+
+        assert len(reviews_by_id_list) == number_of_reviews
+        for position, review in enumerate(reviews_by_id_list):
+            assert review == review_list[position]
--- a/review/models.py
+++ b/review/models.py
@@ -1,3 +1,4 @@
+from django.core.exceptions import ObjectDoesNotExist
 from django.db import models
 from django.core.validators import MinValueValidator, MaxValueValidator
 from dogowner.models import DogOwner
@@ -19,3 +20,10 @@ class Review(models.Model):
         tmp_review.full_clean()
         tmp_review.save()
         return tmp_review
+
+    @staticmethod
+    def get_review_by_daycare_id(daycare_id):
+        try:
+            return Review.objects.filter(daycare_id=daycare_id)
+        except ObjectDoesNotExist:
+            return None


OfirMatasas (approved):
LGTM!

Omeramsc:
I would've liked to see this change in usage on the same PR (meaning the view that lets a user search)

jeniaSakirko:
> I would've liked to see this change in usage on the same PR (meaning the view that lets a user search)

The usage will be in #64 when the DayCare page will have the reviews

review/migrations/0002_test_data.py
```diff
@@ -23,8 +21,7 @@ def generate_data(apps, schema_editor):
 
         with transaction.atomic():
             for review, rating, day_care_id, dog_owner_id in test_data:
-                Review(review=review, rating=rating, daycare_id=DayCare.objects.get(pk=day_care_id),
-                       dogowner_id=DogOwner.objects.get(pk=dog_owner_id)).save()
+                Review.create(review=review, rating=rating, daycare_id=day_care_id, dogowner_id=dog_owner_id)
 
     operations = [
         migrations.RunPython(generate_data),
```
ErezCohenn, line 27: Shouldn't you update the data migration (as you wrote to me in a comment in another PR) ?
And just run migrate so it will generate a new file?

jeniaSakirko: the change is to use the right creation method and not changing the data it self
The reason for this is that the original code didn't run the validations

Omeramsc:
> > I would've liked to see this change in usage on the same PR (meaning the view that lets a user search)
> 
> The usage will be in #64 when the DayCare page will have the reviews, we decided to split this part for separate PR but the change can be part of the original PR

Is that because you work separately on this feature? 
If so, I'll allow it, but I expect future features like this one to include the functionality on the same PR.

jeniaSakirko:
> > > I would've liked to see this change in usage on the same PR (meaning the view that lets a user search)
> > 
> > 
> > The usage will be in #64 when the DayCare page will have the reviews, we decided to split this part for separate PR but the change can be part of the original PR
> 
> Is that because you work separately on this feature? If so, I'll allow it, but I expect future features like this one to include the functionality on the same PR.

@Omeramsc  Cool so in the future this kind of change will be included as part of the main PR
If there are no more comments can you approve / merge?

ErezCohenn (approved):
LGTM

review/tests.py
```diff
@@ -51,3 +51,21 @@ def test_delete_related_dogowner(self, test_review):
         dogowner.delete()
         assert dogowner not in DogOwner.objects.all()
         assert test_review in Review.objects.all()
+
+    def test_get_review_by_daycare_id(self):
+        daycare = DayCare.create(email="review@gmail.com", username="review", password="pass", name="reviewDayCare",
```
Omeramsc, line 56: Unless there's something special about this daycare, we may want to have a daycare fixture under contest on project-level.

review/tests.py
```diff
@@ -51,3 +51,21 @@ def test_delete_related_dogowner(self, test_review):
...
+                                 description="This is test daycare for review", price_per_day=10, capacity=50,
+                                 area="Merkaz", city="Tel-Aviv", address="The best review")
+        reviews = Review.get_review_by_daycare_id(daycare.id)
+        assert len(reviews) == 0
```
Omeramsc, line 60: The function returns None in case there are no reviews, so that's what you should expect here. 
Any reason you check for len?

review/tests.py
```diff
@@ -51,3 +51,21 @@ def test_delete_related_dogowner(self, test_review):
...
+
+        Review.create(review='review 1', rating=1, daycare_id=daycare.id, dogowner_id=1)
```
Omeramsc, line 62: This test uses a different method for daycare and dogowner test data. make it uniform

review/tests.py
```diff
@@ -51,3 +51,21 @@ def test_delete_related_dogowner(self, test_review):
...
+
+        reviews = Review.get_review_by_daycare_id(daycare.id)
+        assert len(reviews) == 1
```
Omeramsc, line 65: Here len is more appropriate, but I'll add another test to check we got the Review we expected (the same one from Review.create)

Omeramsc (changes requested):

review/migrations/0002_test_data.py
```diff
@@ -23,8 +21,7 @@ def generate_data(apps, schema_editor):
 
         with transaction.atomic():
             for review, rating, day_care_id, dog_owner_id in test_data:
-                Review(review=review, rating=rating, daycare_id=DayCare.objects.get(pk=day_care_id),
-                       dogowner_id=DogOwner.objects.get(pk=dog_owner_id)).save()
+                Review.create(review=review, rating=rating, daycare_id=day_care_id, dogowner_id=dog_owner_id)
```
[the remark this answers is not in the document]
jeniaSakirko: Done see #79

review/tests.py
jeniaSakirko: The function return none in case there is ObjectDoesNotExist exception, this search will return empty QuerySet
image

review/tests.py
jeniaSakirko: changed

review/tests.py
jeniaSakirko: fixed

review/tests.py
jeniaSakirko: Added

review/tests.py
Omeramsc: Gotcha

review/tests.py
```diff
@@ -8,12 +8,26 @@
...
+@pytest.fixture
+def test_daycare():
+    return DayCare.create(email="daycare_review@gmail.com", username="daycare_review", password="pass",
+                          name="reviewDayCare", description="This is test daycare for review", price_per_day=10,
+                          capacity=50, area="Merkaz", city="Tel-Aviv", address="The best review")
+
+
+@pytest.fixture
+def test_dogowner():
+    return DogOwner.create(email="dogowner_review@gmail.com", username='dogowner_review', password='pass',
```
Omeramsc, line 25: we are creating those in most models. maybe we can make use of conftest?  (hint: it can work on different levels)

review/tests.py
```diff
@@ -51,3 +65,29 @@ def test_delete_related_dogowner(self, test_review):
         dogowner.delete()
         assert dogowner not in DogOwner.objects.all()
         assert test_review in Review.objects.all()
+
+    def test_get_review_by_daycare_id(self, test_daycare, test_dogowner):
+        reviews = Review.get_review_by_daycare_id(test_daycare.id)
+        assert len(reviews) == 0
+
+        review_list = []
+
```
Omeramsc, line 74: no need for this space

review/tests.py
```diff
@@ -51,3 +65,29 @@ def test_delete_related_dogowner(self, test_review):
...
+        review_list.append(Review.create(review='review 1', rating=1, daycare_id=test_daycare.id,
+                                         dogowner_id=test_dogowner.id))
+
```
Omeramsc, line 77: or this

review/tests.py
```diff
@@ -51,3 +65,29 @@ def test_delete_related_dogowner(self, test_review):
...
+        reviews = Review.get_review_by_daycare_id(test_daycare.id)
+        assert len(reviews) == len(review_list)
+        assert reviews.first() == review_list[0]
+
+        review_list.append(
+            Review.create(review='review 2', rating=2, daycare_id=test_daycare.id, dogowner_id=test_dogowner.id))
+        review_list.append(
```
Omeramsc, line 84: can use `review_list.extend` and add both entries at the same time.

Omeramsc: actually, not only I suggest you to use `extend`, but IMO you should also parametrize this test instead of testing the same method multiple times on the same test.

review/tests.py
```diff
@@ -51,3 +65,29 @@ def test_delete_related_dogowner(self, test_review):
...
+            Review.create(review='review 3', rating=3, daycare_id=test_daycare.id, dogowner_id=test_dogowner.id))
+
+        reviews = Review.get_review_by_daycare_id(test_daycare.id)
+        assert len(reviews) == len(review_list)
+
+        i = 0
```
Omeramsc, line 90: this is not needed. you can use `[enumerate](https://realpython.com/python-enumerate/)`

review/tests.py
jeniaSakirko: > actually, not only I suggest you to use `extend`, but IMO you should also parametrize this test instead of testing the same method multiple times on the same test.

what do you mean?

review/tests.py
jeniaSakirko: I created conftest for the review model and created #80 issue to create one conftest for all models

review/tests.py
jeniaSakirko: I changed to `extend` regarding the parametrize can you share an example of what you thinking about?

review/conftest.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+from daycare.models import DayCare
+from dogowner.models import DogOwner
+from review.models import Review
+
+
+@pytest.fixture
+def test_daycare():
```
Omeramsc, line 8: BTW I will recommend removing the `test_` from all fixture's names. it is a fixture, not a test. it'll also make all tests more readable 
(for a different issue/PR though)

review/tests.py
```diff
@@ -2,38 +2,30 @@
...
-def test_review():
-    return Review.create(review=REVIEW, rating=RATING, daycare_id=DAY_CARE_ID, dogowner_id=DOG_OWNER_ID)
-
-
-@pytest.mark.django_db
+@pytest.mark.django_db()
 class TestReviewModel:
-    def test_persist_review(self, test_review):
-        assert test_review.review == REVIEW
+    def test_persist_review(self, test_review, test_review_data):
```
Omeramsc, line 8: I don't get this test. what is it for?

review/tests.py
```diff
@@ -2,38 +2,30 @@
...
 
-    def test_rating_limit_min(self):
+    def test_rating_limit_min(self, test_review_data):
         with pytest.raises(ValidationError):
-            Review.create(review=REVIEW, rating=-1, daycare_id=DAY_CARE_ID, dogowner_id=DOG_OWNER_ID)
+            Review.create(review=pytest.REVIEW, rating=-1, daycare_id=pytest.DAY_CARE_ID,
+                          dogowner_id=pytest.DOG_OWNER_ID)
 
-    def test_rating_limit_max(self):
+    def test_rating_limit_max(self, test_review_data):
```
Omeramsc, line 22: use parametrize to have `test_rating_limit_min` and `test_rating_limit_max` tested on the same test.
also: is it relevant for this PR?

review/tests.py
```diff
@@ -2,38 +2,30 @@
...
         with pytest.raises(ValidationError):
-            Review.create(review=REVIEW, rating=6, daycare_id=DAY_CARE_ID, dogowner_id=DOG_OWNER_ID)
+            Review.create(review=pytest.REVIEW, rating=6, daycare_id=pytest.DAY_CARE_ID,
+                          dogowner_id=pytest.DOG_OWNER_ID)
```
Omeramsc, line 25: On a different PR/Issue, include this and the former testcase on the same test using parametrize.

review/tests.py
Omeramsc: A test should typically have 3 stages: `arrange`, `act`, `assert`. but here you act -> assert -> arrange -> act ->  assert ->  arrange -> act -> assert.
That's because you're testing 3 different test-cases here (no reviewers, 1 review, 2 reviews).
You can parametrize the test, and so have each of the use cases tested separately; this will keep the test atomic, clean, and readable with only one appearance of each  test case

review/conftest.py
jeniaSakirko: changed

review/tests.py
jeniaSakirko: Old test for testing the data

review/tests.py
jeniaSakirko: changed to parametrize

review/tests.py
```diff
@@ -2,52 +2,49 @@
...
+
+        dogowner_fixture.delete()
+        assert dogowner_fixture not in DogOwner.objects.all()
+        assert review_fixture in Review.objects.all()
+
+    @pytest.mark.parametrize("number_of_reviews", [0, 1, 3])
+    def test_get_review_by_daycare_id(self, number_of_reviews, dogowner_fixture, daycare_fixture):
+        review_list = []
+        for i in range(1, number_of_reviews + 1):
+            review_list.append(Review.create(review=f'review {i}', rating=i, daycare_id=daycare_fixture.id,
```
Omeramsc, line 44: does `rating` and `review` really need to be dynamic? I don't see their values need to be unique (making this usage of range redundant)

review/tests.py
```diff
@@ -2,52 +2,49 @@
...
+                                             dogowner_id=dogowner_fixture.id))
+        reviews = Review.get_review_by_daycare_id(daycare_fixture.id)
+        assert len(reviews) == number_of_reviews
```
Omeramsc, line 47: nit: have an empty line between different parts of the test (arrange, act, assert)

review/tests.py
```diff
@@ -2,52 +2,49 @@
...
+
+        for position, review in enumerate(review_list):
+            assert review == review_list[position]
```
Omeramsc, line 50: On a second look, I don't get this assertion. why are we checking that `review_list[n]==review_list[n]`?

review/conftest.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+from daycare.models import DayCare
+from dogowner.models import DogOwner
+from review.models import Review
+
+
+@pytest.fixture
+def daycare_fixture():
```
Omeramsc, line 8: It is known that this is a fixture, it is written right above it. :) `daycare` should be enough

review/tests.py
jeniaSakirko: changed

review/tests.py
jeniaSakirko: added

review/tests.py
jeniaSakirko: lol nice catch!
fixed

review/conftest.py
jeniaSakirko: changed

Yuval-Vino (approved):
LGTM!

review/tests.py
```diff
@@ -2,52 +2,50 @@
...
         assert dogowner not in DogOwner.objects.all()
-        assert test_review in Review.objects.all()
+        assert review in Review.objects.all()
+
+    @pytest.mark.parametrize("number_of_reviews", [0, 1, 3])
+    def test_get_review_by_daycare_id(self, number_of_reviews, dogowner, daycare):
+        review_list = []
+        for i in range(1, number_of_reviews + 1):
+            review_list.append(Review.create(review='review', rating=1, daycare_id=daycare.id,
+                                             dogowner_id=dogowner.id))
```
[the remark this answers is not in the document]
jeniaSakirko: you mean with `parametrize`?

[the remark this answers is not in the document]
jeniaSakirko: Although I think this approach is more readable fixed :)

ormergi (changes requested):
HI Jenia, 
No need to mention tests in the commit message, please remove it.
Next time make sure to separate refactoring/formatting changes to their own commits or PR,
its hard to review this one as there is a lot of noise on the diff.

OfirMatasas (approved):
Looks like everything was mentioned and fixed by now :)
LGTM!

tamirmatok (approved):
LGTM

review/tests.py
```diff
@@ -2,52 +2,48 @@
...
         dogowner.delete()
         assert dogowner not in DogOwner.objects.all()
-        assert test_review in Review.objects.all()
+        assert review in Review.objects.all()
+
+    @pytest.mark.parametrize("number_of_reviews", [0, 1, 3])
+    def test_get_review_by_daycare_id(self, number_of_reviews, dogowner, daycare):
+        review_list = [Review.create(review='review', rating=1, daycare_id=daycare.id, dogowner_id=dogowner.id)
+                       for _ in range(number_of_reviews)]
+        reviews_response = Review.get_review_by_daycare_id(daycare.id)
```
Omeramsc, line 44: It's a nitpick, but still important to understand:
to improve test readability, the spaces should divide the arrange, act, and assert parts of the test. 
the `review_list` is the arrange, the `reviews_response` is the act (it uses the method this test is made for) and the rest is the assert. 

p.s: long parts can use spaces as well. but this is irrelevant to this test

Omeramsc, line 44: why is it called `reviews_response`?

Omeramsc (changes requested):
some last minor changes. let's merge it soon :)

ormergi (approved):
LGTM

Omeramsc (approved):
LGTM

PR #84, opened 2022-04-20
Create static reviews for all DayCares

### LINKED ISSUE #83: Add static reviews
## Expected Behavior

Because the project uses only static users without the ability to register at the moment, there is a need for a large number of static reviews for daycares for regular work.

## Current Behavior

There are only 3 static reviews in the data migrations

## Acceptance Criteria

New migration was added to create static reviews for all DayCares

jeniaSakirko:
Signed-off-by: Jenia Sakirko <Jenia.Sakirko@cyberark.com>

### Desired Outcome

Reviews are created automatically for all DayCares

### Implemented Changes

New migration that create all reviews for all daycares

### Connected Issue/Story

Resolves #83 

### Definition of Done

New migration was added to create static reviews for all DayCares

#### Changelog

  CHANGELOG update

#### Test coverage

- [ ] This PR includes new unit and integration tests to go with the code
  changes
- [x] The changes in this PR do not require tests

#### Documentation

### DIFF
--- a/review/migrations/0003_static_reviews_data.py
+++ b/review/migrations/0003_static_reviews_data.py
@@ -1,0 +1,55 @@
+from django.db import migrations, transaction
+import random
+import textwrap
+
+REVIEW_AND_RATE_LIST = [
+    (textwrap.dedent("""The dog supervisors are awesome people and love working with dogs. Everyone I have met there is
+    so friendly and warm. My dog LOVES them and is so happy to go there each morning and I am happy that we have a
+    place where he can be cared for during the day when I am gone with people that he cares about and who care for
+    him!"""), 5),
+    (textwrap.dedent("""We left our small dog Neo at Manoir Kanisha for a weekend and could not be more impressed with
+    the treatment he received. Every imaginably detail is anticipated and attended by Nicole and her caring staff.
+    The facilities are of luxury hotel calibre, and every effort is made to make your pet’s stay a pleasant and calm
+    one. Nicole even took some photos of Neo with my 6-year old daughter so that i could show my dad, Neo’s actual
+    owner. We’re impressed beyond words, and enthusiastically recommend Kanisha to others. Thanks so much!"""), 4),
+    (textwrap.dedent("""We brought our dog to Manoir Kanisha when it first opened in 1990. I was worried about him
+    while on vacation so I called Nicole. She reassured me that he was doing very well and she was enhancing his meals
+    with burger and rice! Kanisha is not just a place to board your dog, it’s a five star doggy hotel. Nicole and her
+    staff are so very knowledgeable – and kind, caring and loving that, since 1990, I haven’t even thought about taking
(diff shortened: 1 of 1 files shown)


ErezCohenn (commented):
I recommend to you take a look at how I implemented the static daycare users,
without using random, and using `list comprehension`.
You can take a look [here](https://github.com/redhat-beyond/Hotails/pull/690) the comments that omer left about it.

jeniaSakirko:
> I recommend to you take a look at how I implemented the static daycare users, without using random, and using `list comprehension`. You can take a look [here](https://github.com/redhat-beyond/Hotails/pull/690) the comments that omer left about it.

IMO list comprehension when running nested loop is not so readable but its looks cleaner so I changed it, kept the random so we will have different reviews in each daycare

OfirMatasas (approved):
I agree with Jenia about list comprehension readability. One-lining isn't always the best option.
Besides that, great job!

Yuval-Vino:
> > I recommend to you take a look at how I implemented the static daycare users, without using random, and using `list comprehension`. You can take a look [here](https://github.com/redhat-beyond/Hotails/pull/690) the comments that omer left about it.
> 
> IMO list comprehension when running nested loop is not so readable but its looks cleaner so I changed it, kept the random so we will have different reviews in each daycare

I have to agree that list comprehension with double for is a bit unreadable, I prefer the previous version with the append method,  anyway LGTM!

Yuval-Vino (approved):

Omeramsc:
Are those static reviews used (or will be used later) for tests? if so, using random is out of the question, as those need to be reliable and explicit.
(otherwise its fine)

jeniaSakirko:
> Are those static reviews used (or will be used later) for tests? if so, using random is out of the question, as those need to be reliable and explicit. (otherwise its fine)
> 
> P.S: I see all the reviews got a space at the beginning. might want to just add the space in the loop instead.

Its not for the tests (we already have reviews for the tests https://github.com/redhat-beyond/Hotails/blob/main/review/migrations/0002_test_data.py) the purpose of this data is to see better in the UI when using static users

ErezCohenn (approved):
LGTM

Omeramsc:
> > Are those static reviews used (or will be used later) for tests? if so, using random is out of the question, as those need to be reliable and explicit. (otherwise its fine)
> > P.S: I see all the reviews got a space at the beginning. might want to just add the space in the loop instead.
> 
> Its not for the tests (we already have reviews for the tests https://github.com/redhat-beyond/Hotails/blob/main/review/migrations/0002_test_data.py) the purpose of this data is to see reviews in the UI when checking DayCare page
> 
> Regarding the space please notice that there are four reviews and the spaces are due to new line, e.g. no space in begging of the review

You're right, 
as it is a bit confusing - try using `dedent` from the `textwrap` library

Omeramsc (approved):
LGTM

PR #89, opened 2022-04-27
Add search day-care functionality to dog-owner homepage.

### LINKED ISSUE #90: Add search day-care functionality to dog-owner homepage.
## Expected Behavior

Please describe the behavior you are expecting
The dog-owner user will search for day-care that relevant for him.
The search will filter by "Area" , "City" and "Max price per day".
If the user didnt search for day-care he will see all Day-Cares in dog-owner homepage.

## What is the current behavior?
dog-owner homepage present all day-cares without filters.

### Steps to Reproduce

1. Extend dog_owner_home view and send compatible context.
2. Create form for the search.
3. Delete "search" button from navbar.
4. Extend dog_owner_homepage.html and present the search form.
5. Cover with tests.

tamirmatok:
### Desired Outcome

The dog-owner user will search for day-care that relevant for him.
If the user didnt search for day-care he will see all Day-Cares in dog-owner homepage.

### Implemented Changes

- Adding search functionality to dog-owner homepage.
- Adding "area" choices-class to day-care model for make the data model compatible and to improve the search.
- Delete "search" button from navbar.

### Connected Issue/Story

depends on #95
Closes #90 

#### Test coverage

  changes
- [ ] The changes in this PR do not require tests

#### Documentation

Capture3


main/tests.py
```diff
@@ -87,3 +88,44 @@ def test_unlogged_user_access_to_homepage(self, client):
...
+                       'start_date': "25/05/2022",
+                       'end_date': "28/05/2022",
+                       }
+        response = client.post('/homepage/', search_form, follow=True)
+        day_care_queryset = response.context['day_care_queryset']
+        for day_care in DayCare.objects.all():
+            if day_care.price_per_day <= 800 and day_care.area.lower() == 'center'\
+                    and day_care.city.lower() == 'tel aviv':
+                assert day_care in day_care_queryset
+
```
tamirmatok, line 116: After #95 will approve dates check will be added .

daycare/migrations/0006_alter_daycare_area.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 4.0.3 on 2022-04-27 13:37
```
Yuval-Vino, line 1: do we need this file?

daycare/test_image.py
```diff
@@ -20,3 +22,13 @@ def test_image_creation_with_invalid_image_url(self, create_daycare_user):
...
+
+    def test_daycare_has_customized_profile_image(self, create_image1, create_image2, create_daycare_user):
+        daycare_profile_image = create_daycare_user.get_daycare_primary_image_url()
+        assert daycare_profile_image != DEFAULT_DAYCARE_PROFILE_URL
+        assert daycare_profile_image is not None
+
+    def test_daycare_has_default_profile_image(self, create_daycare_user):
+        daycare_profile_image = create_daycare_user.get_daycare_primary_image_url()
+        assert daycare_profile_image == DEFAULT_DAYCARE_PROFILE_URL
+        assert daycare_profile_image is not None
```
Yuval-Vino, line 34: I think this test is redundant , if profile_image = default so obviously its not None

main/tests.py
```diff
@@ -87,3 +88,44 @@ def test_unlogged_user_access_to_homepage(self, client):
...
         assert response['Location'] == '/login/?next=/homepage/'
+
+
+@pytest.mark.django_db
+class TestDogOwnerHomePageView:
+    def test_dog_owner_present_all_day_cares_at_entrypoint(self, client, create_dog_owner_user):
+        client.force_login(user=create_dog_owner_user.user)
+        response = client.get("/homepage/")
+        day_care_queryset = response.context['day_care_queryset']
+        assert day_care_queryset.count() == DayCare.objects.all().count()
```
Yuval-Vino, line 99: maybe instead of count use set(day_care_queryset) == set(daycare.objects.all) because if the count is the same we are not sure we presenting all daycares, maybe we present the same one 30 times?

main/tests.py
```diff
@@ -87,3 +88,44 @@ def test_unlogged_user_access_to_homepage(self, client):
...
+                       'name': "",
+                       'start_date': "25/05/2022",
+                       'end_date': "28/05/2022",
+                       }
+        response = client.post('/homepage/', search_form, follow=True)
+        day_care_queryset = response.context['day_care_queryset']
+        for day_care in DayCare.objects.all():
+            if day_care.price_per_day <= 800 and day_care.area.lower() == 'center'\
+                    and day_care.city.lower() == 'tel aviv':
+                assert day_care in day_care_queryset
```
Yuval-Vino, line 115: maybe also check that all day_care_queryset is in filter range, 
your test checks that all filtered daycares are shown, maybe add a test that all daycares in day_care_queryset are the correct ones.
Easy hit append all daycares in the for to list, then assert equality between this list to day_care_queryset

daycare/migrations/0006_alter_daycare_area.py
tamirmatok: This file auto-created by running makemigration after adding choices to area field.

daycare/test_image.py
tamirmatok: please note that this comment related to #67

daycare/migrations/0006_alter_daycare_area.py
Yuval-Vino: maybe add to 0001 initial the specific field, instead of generating a new file for this
tell me what you think?

main/tests.py
tamirmatok: Im not sure I fully understand , but in this test we want to check that day_care_queryset contain all Daycare objects when we are entering to dog_owner homepage.
Im not sure that "set" is define on queryset and since every user is unique it is promise that every object in the queryset is diffrent from the other.

main/tests.py
tamirmatok: You right . I guess we can use a counter in the for loop instead and to check in the end if counter = day_care_queryset.count.

main/tests.py
Yuval-Vino: that's right for the right side of the assert (daycare.objects), what about what is presented in HTML (daycare_queryset), how can you be sure that they are unique and don't have dups, for example, daycare shown twice in HTML or things like that

Yuval-Vino: And anyway it just seems weird to conclude equality between sets by their size, even if all the values are unique

daycare/migrations/0006_alter_daycare_area.py
tamirmatok: https://github.com/redhat-beyond/Hotails/pull/65#discussion_r847086629 :bowtie:

Yuval-Vino: Yep, thank you for notifying me in this 🙃

main/tests.py
tamirmatok: The HTML running for loop and create a card for each daycare in day_care query set so thats way I think is enough to test **what this queryset contatin** .
Since day_care_query is **subset** of all day_cares queryset, that means that if the number of elements between them is equal they are have to contatin the same values .

tamirmatok: But if u think it is less readble I will try your segestion about "set(day_care_queryset) == set(daycare.objects.all)" (:

ErezCohenn:
@tamirmatok this PR is ready for review? notice that the PR is on draft

tamirmatok:
> @tamirmatok this PR is ready for review? notice that the PR is on draft

Now it is (:

OfirMatasas:
The idea about making the area a choice box is great!
IMO, I think we should do the same with the cities, to avoid a situation where the daycare is on "Tel-Aviv", but the dog owner searched for "Tel aviv" or "Tel Aviv - Yaffo".
What do you think? @tamirmatok

ErezCohenn (commented):
What happens in the case that after filtering the daycares, there were no daycares that match the selection of the user? I think in that case we should write to the user something like:
#### There were no daycares found that match the filtering that was selected

tamirmatok:
> What happens in the case that after filtering the daycares, there were no daycares that match the selection of the user? I think in that case we should write to the user something like:
> 
> #### There were no daycares found that match the filtering that was selected

We have a generic sentence for all the search results : "Found %d results for your dog" . How does it sound ? @ErezCohenn

ErezCohenn:
> > What happens in the case that after filtering the daycares, there were no daycares that match the selection of the user? I think in that case we should write to the user something like:
> > #### There were no daycares found that match the filtering that was selected
> 
> We have a generic sentence for all the search results : "Found %d results for your dog" . How does it sound ? @ErezCohenn

Sounds good :)

tamirmatok:
> The idea about making the area a choice box is great! IMO, I think we should do the same with the cities, to avoid a situation where the daycare is on "Tel-Aviv", but the dog owner searched for "Tel aviv" or "Tel Aviv - Yaffo". What do you think? @tamirmatok

Hi @OfirMatasas , first of all I think you right and that issue appeared in my head as well.
Let me share with you my thoughts about it .
First, please note that "search" data fields/choices and "user-creation" fields/choices kind of come together(In our case search/create-user that involves cities field).
I think that to add city a choice like you mention indeed will work in "static world" with no registration,
but in that world we can control the cities input and just avoid mistakes like you mention. 

I think the right approch to deal with this issue is kind of "preapre the ground" for supporting in user-registration, with django-cities(see https://pypi.org/project/django-cities/), it is look better to me because of the benefits of global support or even local but with more percise data.
I believe we should meet as a team and to decide if we want to dedicate time on this kind of stuff at this point of the course.

OfirMatasas:
> > The idea about making the area a choice box is great! IMO, I think we should do the same with the cities, to avoid a situation where the daycare is on "Tel-Aviv", but the dog owner searched for "Tel aviv" or "Tel Aviv - Yaffo". What do you think? @tamirmatok
> 
> Hi @OfirMatasas, I think you right and that issue appeared in my head as well. Let me share with you my thoughts about it . First, please note that "search" data fields/choices and "user-creation" fields/choices kind of come together(In our case search/create-user that involve city choices field). I think that to add city a choice like you mention indeed will work in "static world" with no user-registration, but in that world, we can control the cities input and just avoid mistakes like that.
> 
> I think the right approch to deal with this issue is kind of "preapre the ground" for supporting in user-registration, with django-cities(see https://pypi.org/project/django-cities/), it looks better to me because of the benefits of global support or even local but with more percise data. I believe we should meet as a team and to decide if we want to dedicate time on this kind of stuff at this point of the course.

I think you're right.
Besides that, switching the cities to choices will effect the rest of the project, so we'll discuss this on our next meeting.

ErezCohenn (approved):
LGTM

daycare/forms.py
```diff
@@ -0,0 +1,31 @@
+from django import forms
+from .models import DayCare, Area
+from django.db.models import Max, Min
+
+
+class DayCareSearchForm(forms.Form):
+    min_price = DayCare.objects.all().aggregate(Min('price_per_day')).get('price_per_day__min')
+    max_price = DayCare.objects.all().aggregate(Max('price_per_day')).get('price_per_day__max')
```
Omeramsc, line 8: I'm not sure the user should be aware of the min\max values a daycare can use for his pricing

daycare/forms.py
```diff
@@ -0,0 +1,31 @@
...
+    start_date = forms.DateField(required=True, label='  Start date',
+                                 widget=forms.widgets.DateInput(attrs={'class': 'form-control', 'type': 'date'}))
+    end_date = forms.DateField(required=True, label='  End date',
+                               widget=forms.widgets.DateInput(attrs={'class': 'form-control', 'type': 'date'}))
+
+    area = forms.ChoiceField(required=False, label='Area',
+                             choices=(Area.choices + [('', 'All'), ]), initial="")
+    city = forms.CharField(required=False, label='City')
```
Omeramsc, line 16: Don't we have a list of cities? a free-text would be tough.

daycare/models.py
```diff
@@ -10,13 +10,19 @@
...
+
+
 class DayCare(models.Model):
     user = models.OneToOneField(User, on_delete=models.CASCADE, default=None, blank=True, null=False, editable=True)
     name = models.CharField(max_length=20, blank=True, unique=True, validators=[MaxLengthValidator])
     description = models.TextField(blank=True, null=True)
     price_per_day = models.IntegerField(blank=False, null=False, default=0)
     capacity = models.IntegerField(null=False, blank=True)
-    area = models.CharField(max_length=20, blank=True, validators=[MaxLengthValidator])
+    area = models.CharField(max_length=20, blank=True, validators=[MaxLengthValidator], choices=Area.choices)
```
Omeramsc, line 25: Shouldn't we include that in some of the daycare fixtures?

dogowner/views.py
```diff
@@ -1,11 +1,32 @@
...
+            filter_day_cares = DayCare.objects.filter(area__startswith=form['area'].value(),
+                                                      city__icontains=form['city'].value(),
+                                                      name__icontains=form['name'].value(),
+                                                      price_per_day__lte=form['price_per_day'].value())
+
+            available_day_cares = Order.get_all_day_cares_available_on_dates(form['start_date'].value(),
+                                                                             form['end_date'].value())
+            day_care_queryset = filter_day_cares.intersection(available_day_cares)
+
+            context = {
```
Omeramsc, line 27: no need to include it twice; since both use the same parameters, just include it once after the conditional statements.

main/tests.py
```diff
@@ -76,4 +78,59 @@ def test_dog_owner_homepage_is_visible_for_dog_owner(self, client, create_dog_ow
...
         response = client.get("/homepage/")
         assert response.status_code == 200
-        assert list(response.context['daycares']) == list(DayCare.objects.all())
+        assert set(response.context['day_care_queryset']) == set(DayCare.objects.all())
+
+
+@pytest.mark.django_db
+class TestDogOwnerHomePageView:
+    def test_search_daycare_present_only_available_daycares_on_specific_dates(self, client, create_dog_owner_user):
+        test_day_care = DayCare.create(username='testuser', email='test@email.com', password='test_password',
```
Omeramsc, line 87: didn't we add a bunch of fixtures to conftest so we won't need this anymore?

orders/tests.py
```diff
@@ -52,3 +52,14 @@ def test_dog_owner_id_is_deleted_when_dog_owner_is_deleted(self, create_order):
     def test_daycare_id_is_deleted_when_daycare_is_deleted(self, create_order):
         DayCare.objects.get(id=1).delete()
         assert Order.objects.get(id=create_order.id).daycare_id is None
+
+    def test_get_all_day_cares_available_on_dates_function(self):
+        start_date = "2022-05-02"
+        end_date = "2022-08-02"
+        day_care = DayCare.create(username='testuser', email='test@email.com', password='test_password',
```
Omeramsc, line 59: didn't we add a bunch of fixtures to conftest so we won't need this anymore?

orders/tests.py
```diff
@@ -52,3 +52,14 @@ def test_dog_owner_id_is_deleted_when_dog_owner_is_deleted(self, create_order):
     def test_daycare_id_is_deleted_when_daycare_is_deleted(self, create_order):
         DayCare.objects.get(id=1).delete()
         assert Order.objects.get(id=create_order.id).daycare_id is None
+
+    def test_get_all_day_cares_available_on_dates_function(self):
```
Omeramsc, line 56: redundant "function"

orders/tests.py
```diff
@@ -52,3 +52,14 @@ def test_dog_owner_id_is_deleted_when_dog_owner_is_deleted(self, create_order):
...
+        start_date = "2022-05-02"
+        end_date = "2022-08-02"
```
Omeramsc, line 58: seems like a classic case for a `parametrize` test to me.

orders/tests.py
```diff
@@ -52,3 +52,14 @@ def test_dog_owner_id_is_deleted_when_dog_owner_is_deleted(self, create_order):
...
+        day_care = DayCare.create(username='testuser', email='test@email.com', password='test_password',
+                                  name='name', description='test description', price_per_day=500,
+                                  capacity=1, area='Center', city='Tel_aviv', address='address')
+        assert day_care in Order.get_all_day_cares_available_on_dates(start_date, end_date)
+        Order.create(dog_owner_id=DogOwner.objects.get(id=1), daycare_id=day_care,
+                     start_date="2022-05-02", end_date="2022-08-02", price_per_day=500).approve_order()
+        assert day_care not in Order.get_all_day_cares_available_on_dates(start_date, end_date)
```
Omeramsc, line 65: This doesn't really belong here. instead, you should add a test that ensure a specific daycare with an order created for it is not found by the function.

orders/models.py
```diff
@@ -71,3 +73,38 @@ def cancel_order(self):
 
     def get_order_total_price(self):
         return self.price_per_day * (self.end_date - self.start_date).days
+
+    @staticmethod
+    def get_capacity_of_daycare_in_dates_range(daycare_id, start_date, end_date):
+        relevant_orders = Order.objects.filter(daycare_id=daycare_id, status__in=['A', 'O'])
+        start_date = datetime.date(year=start_date.year, month=start_date.month, day=start_date.day)
```
Omeramsc, line 80: why do we declare start_date again? since we don't use python's `Typing` it is a bit confusing. maybe formatted_start_date will be better.

orders/models.py
```diff
@@ -71,3 +73,38 @@ def cancel_order(self):
...
+        end_date = datetime.date(year=end_date.year, month=end_date.month, day=end_date.day)
```
Omeramsc, line 81: same

orders/models.py
```diff
@@ -71,3 +73,38 @@ def cancel_order(self):
 
     def get_order_total_price(self):
         return self.price_per_day * (self.end_date - self.start_date).days
+
+    @staticmethod
+    def get_capacity_of_daycare_in_dates_range(daycare_id, start_date, end_date):
```
Omeramsc, line 78: There is a LOT of logic here. how does it not have a test?

orders/models.py
```diff
@@ -71,3 +73,38 @@ def cancel_order(self):
...
+                elif current_date > end_date:
+                    break
+                else:
+                    capacity_per_day_list[day] = capacity_per_day_list[day] + 1
+
+        return capacity_per_day_list
+
+    @staticmethod
+    def get_all_day_cares_available_on_dates(start_date: str, end_date: str) -> QuerySet:
+        start_date = datetime.date.fromisoformat(start_date)
```
Omeramsc, line 103: same as I commented on the former function.

main/tests.py
```diff
@@ -76,4 +78,59 @@ def test_dog_owner_homepage_is_visible_for_dog_owner(self, client, create_dog_ow
...
+                       'start_date': "2022-05-03",
+                       'end_date': "2022-05-08",
+                       }
+        response = client.post('/homepage/', search_form, follow=True)
+        day_care_queryset = response.context['day_care_queryset']
+        available_day_cares = Order.get_all_day_cares_available_on_dates("2022-05-03", "2022-05-08")
+        filters_day_cares = DayCare.objects.filter(area__startswith='C',
+                                                   city__icontains="tel aviv",
+                                                   price_per_day__lte=800)
+        assert set(day_care_queryset) == set(available_day_cares.intersection(filters_day_cares))
```
Omeramsc, line 121: Do we test the functionality of the filter? I mean: do we expect specific search results? what are the day cares that exists during this test?

main/tests.py
```diff
@@ -76,4 +78,59 @@ def test_dog_owner_homepage_is_visible_for_dog_owner(self, client, create_dog_ow
...
+                       'city': "",
+                       'price_per_day': 100,
+                       'name': "",
+                       'start_date': "2022-05-03",
+                       'end_date': "2022-05-08",
+                       }
+        response = client.post('/homepage/', search_form, follow=True)
+        day_care_queryset = response.context['day_care_queryset']
+        assert test_day_care in day_care_queryset
+        Order.create(dog_owner_id=DogOwner.objects.get(id=1), daycare_id=test_day_care,
```
Omeramsc, line 101: belong in a different test

main/tests.py
```diff
@@ -76,4 +78,59 @@ def test_dog_owner_homepage_is_visible_for_dog_owner(self, client, create_dog_ow
         client.force_login(user=create_dog_owner_user.user)
         response = client.get("/homepage/")
         assert response.status_code == 200
-        assert list(response.context['daycares']) == list(DayCare.objects.all())
+        assert set(response.context['day_care_queryset']) == set(DayCare.objects.all())
```
Omeramsc, line 81: If this test is ensuring that the page will have a list of all the day cares, maybe it should be mentioned in its name. or in a different test.

Omeramsc (changes requested):

daycare/forms.py
tamirmatok: Please see https://github.com/redhat-beyond/Hotails/pull/89#issuecomment-1120235525 and let me know what you think.

main/tests.py
tamirmatok: The reason I didnt use the fixture from the confest is because this is a speacil case of day-care with capcity=1.

orders/models.py
tamirmatok: I have mentioned in the commit message "get_capacity_of_daycare_in_dates_range" function took from https://github.com/redhat-beyond/Hotails/pull/95 .
waiting for approval." , and in the PR description that this PR depends on #95.
After #95 will merge test implementation should be there as well .

orders/tests.py
tamirmatok: see https://github.com/redhat-beyond/Hotails/pull/89/commits/c3ef4074998c115f486216beefc1b26d73cd4d97#r874035611 and let me know what you think (:

orders/models.py
tamirmatok: Sorry if it wasn't clear that this function took from #95, I guess you already mention this comment there.

orders/models.py
tamirmatok: NP .

main/tests.py
tamirmatok: NP .

main/tests.py
tamirmatok: The day-cares that exists during this test are from the migration file . 
In this test I considered the filters queryset as the expected results and make sure that the 'day_care_queryset' from the response is the same .

daycare/forms.py
tamirmatok: I guess you right .

dogowner/views.py
tamirmatok: You right NP.

main/tests.py
tamirmatok: Can you explain ? Do you think that create day-care with capcity=0, and test that it doesn't appear in the queryset is better approch ?

daycare/forms.py
Omeramsc: Can you sum it up? I'm not sure I'm following. Anyway free-text can lead to issues (e.g: פתח-תקוה, פתח תקווה, פתח-תקווה, פתח תקוה)

main/tests.py
Omeramsc: What is the difference? how much more work do you need to do if you use the fixture?

main/tests.py
Omeramsc: on the first assertion you check that a daycare with capacity is available, 
in the second one you check  that a daycare with no capacity is not available. 
those are two different cases.
BUT, on second thought, I think it can keep it like this. but put spaces between different parts of the test.

Omeramsc: and change the name of the test. it doesn't check that the daycare is available  on specific dates. it checks that a daycare is not available after its capacity been *reduced* to 0.

main/tests.py
tamirmatok: Just one for loop for decreasing the capcity from 50 to 0 (that means to create 50 orders to fulfill the capacity ) 
Another option is to change the capacity in the fixture . @Omeramsc

main/tests.py
tamirmatok: You right thanks for making it clear.

main/tests.py
Omeramsc: The test should verify the functionality would work in any case, and not only a specific one we carefully created. 

what is `intersection`? what does it provide? This proves the usefulness of declaring the variables before asserting them...

daycare/forms.py
tamirmatok: List of cities seems to me good solution for static-users ,but because all users are static we can control the cities input and just avoid mistakes like that.
If we will want to support registration in the future it might limit us . 
I saw django-cities(see https://pypi.org/project/django-cities/), it seems to have benefits of global support or even local but with more percise data.
I have no problem to create list of cities just want to know your opinion about it . @Omeramsc

orders/tests.py
Omeramsc: either:
1. use the one in the contest and deal with the difference
2. add one with capacity=1  (my least favorite...)
3. use fixture-factory to create it dynamically.

orders/tests.py
Omeramsc: no. the test is named `test_get_all_day_cares_available_on_dates_function`.
a `not in` assertion test something completely different - it tests that a daycare is *not available*

main/tests.py
tamirmatok: NP I will make it more clear .Since we want to filter by capacity of day-care in specific dates we have to create a query set that contain all day-cares avaible in dates , and the second one according to the other filters of the search(we have django built in filter for that ).
The intersection between those querysets will give us all day-cares avaible in specific dates that fit with the other filters .

Do you have an idea how can I make this test more general like you mention ?
Because if I will try all the combination of the filters we won't get out of it. @Omeramsc

orders/tests.py
tamirmatok: Got your point.

daycare/models.py
tamirmatok: Area field in create_day_care_user from the confest been updated.

dogowner/views.py
tamirmatok: Fixed.

main/tests.py
tamirmatok: Fixed.

main/tests.py
tamirmatok: Fixed

main/tests.py
tamirmatok: Fixed by changing the name to " test_dog_owner_homepage_present_all_daycares"

daycare/forms.py
tamirmatok: Fixed

orders/tests.py
tamirmatok: Fixed with option 1.

orders/tests.py
tamirmatok: Fixed.

daycare/forms.py
ormergi: The UI is an entry point for this app and as such it should perform validation on the input,
regardless if the users are static or not.
Nothing prevents for the next dev to create another static user and enter non valid city name, right?
If this is a lot of work you can do it on a follow up PR and write in the PR description about it.

main/tests.py
```diff
@@ -72,8 +74,66 @@ def test_unlogged_user_access_to_homepage(self, client):
...
+                       'price_per_day': 100,
+                       'name': "",
+                       'start_date': "2022-05-03",
+                       'end_date': "2022-05-08",
+                       }
+        daycare_user = create_daycare_user
+        client.force_login(user=create_dog_owner_user.user)
+        response = client.post('/homepage/', search_form, follow=True)
+        day_care_queryset = response.context['day_care_queryset']
+        assert daycare_user in day_care_queryset
```
ormergi, line 100: It would make the test look nicer if the assertion would be separated with newline from the test setup logic (all over this file).

main/tests.py
```diff
@@ -72,8 +74,66 @@ def test_unlogged_user_access_to_homepage(self, client):
...
+        response = client.post('/homepage/', search_form, follow=True)
+        day_care_queryset = response.context['day_care_queryset']
+        available_day_cares = Order.get_all_day_cares_available_on_dates("2022-05-03", "2022-05-08")
+        filter_day_cares = DayCare.objects.filter(area__startswith='C',
+                                                  city__icontains="tel aviv",
+                                                  price_per_day__lte=800)
+        expected_queryset = available_day_cares.intersection(filter_day_cares)
+        assert set(day_care_queryset) == set(expected_queryset)
+
+    def test_search_for_day_care_with_start_date_greater_than_end_date_show_error_and_present_all_daycares(
```
ormergi, line 126: By mentioning `with_start_date_greater_than_end_date` it makes me think this is the only case all daycares will show.
Would it make sense to rename this test to something like:
`test_dog_owner_page_view_presents_all_daycares_when_search_results_is_empty`?
In other words, would all daycare show in case of a search error?

ormergi (commented):
Hi Tamir, 
Note that there is no need to add `.` at the end of a commit message title, please remove them.
There is typo at second commit title `funconallity` instead of `functionality`.
Please see my inline comments

PR #95, opened 2022-04-28
Creating orders.html, presenting orders history

### LINKED ISSUE #94: Create orders page, filled with user's orders history
## Expected Behavior

Showing user's orders history on one page, approachable from the navbar, where the user can cancel upcoming orders, and approve pending orders as a daycare (when it makes sense).

## Current Behavior

User's orders history is not presented to the user

OfirMatasas:
A page on Hotails, where the user can watch the history of all of his
orders, cancel the orders which havn't started yet, and approve orders
as a daycare.

### Desired Outcome

A page on Hotails, where the user can watch the history of all of his orders, cancel the orders which haven't started yet, and approve orders as a daycare.

### Implemented Changes

- Create orders.html + orders.css files
- Added logic in getting to the orders page and presenting orders information (base_template navbar, main.urls, orders.views)
- added 3 methods in Orders:
    1. Check if the order is approvable
    2. Check if the order is cancelable
    3. Get the status of the order 

### Connected Issue/Story

Closes #94

### Definition of Done

All the options mentioned in the desired outcome are available to the users.

### Picture of the orders page
Untitled

### DIFF
--- a/main/tests.py
+++ b/main/tests.py
@@ -1,4 +1,5 @@
 import pytest
+from orders.models import Order
 from daycare.models import DayCare
 
 
@@ -77,3 +78,28 @@ class TestHomepageView:
         response = client.get("/homepage/")
         assert response.status_code == 200
         assert list(response.context['daycares']) == list(DayCare.objects.all())
+
+
+@pytest.mark.django_db
+class TestOrdersView:
+    def test_only_orders_which_belongs_to_user_as_daycare_are_displayed(self, client, create_daycare_user):
+        client.force_login(user=create_daycare_user.user)
+        response = client.get("/orders/")
+
+        response_orders_list = list(response.context['orders'])
+        all_orders_of_daycare_list = list(Order.objects.filter(daycare_id=create_daycare_user))
+        response_user = response.context['user']
+
+        assert response_user == 'daycare'
+        assert response_orders_list == all_orders_of_daycare_list
+
+    def test_only_orders_which_belongs_to_user_as_dog_owner_are_displayed(self, client, create_dog_owner_user):
+        client.force_login(user=create_dog_owner_user.user)
+        response = client.get("/orders/")
+
+        response_orders_list = list(response.context['orders'])
+        all_orders_of_dog_owner_list = list(Order.objects.filter(dog_owner_id=create_dog_owner_user))
+        response_user = response.context['user']
+
+        assert response_user == 'dog_owner'
+        assert response_orders_list == all_orders_of_dog_owner_list
--- a/orders/models.py
+++ b/orders/models.py
@@ -2,6 +2,7 @@ from django.db import models
 from django.utils import timezone
 from daycare.models import DayCare
 from dogowner.models import DogOwner
+import datetime
 
 
 class StatusOptions(models.TextChoices):
@@ -71,3 +72,51 @@ class Order(models.Model):
 
     def get_order_total_price(self):
         return self.price_per_day * (self.end_date - self.start_date).days
+
+    def get_order_status(self):
+        if self.status == StatusOptions.Pending:
+            return "Pending"
+        elif self.status == StatusOptions.Approved:
+            return "Approved"
+        elif self.status == StatusOptions.Canceled:
+            return "Canceled"
+        elif self.status == StatusOptions.OnGoing:
+            return "Ongoing"
+        return "Finished"
+
+    @staticmethod
+    def get_capacity_of_daycare_in_dates_range(daycare_id: DayCare, start_datetime: datetime, end_datetime: datetime):
+        number_of_days = (end_datetime - start_datetime).days
+        capacity_per_day: list[int] = [0] * number_of_days
+        start_date = datetime.date(year=start_datetime.year, month=start_datetime.month, day=start_datetime.day)
+        end_date = datetime.date(year=end_datetime.year, month=end_datetime.month, day=end_datetime.day)
+        relevant_orders = Order.objects.filter(daycare_id=daycare_id,
+                                               status__in=['A', 'O'],
+                                               end_date__gte=start_date,
+                                               start_date__lte=end_date)
+
+        for order in relevant_orders:
+            for day in range(number_of_days):
+                current_date = order.start_date + datetime.timedelta(days=day)
+                if current_date < start_date:
+                    continue
+                elif current_date > end_date:
+                    break
+                else:
+                    capacity_per_day[day] = capacity_per_day[day] + 1
+
+        return capacity_per_day
+
+    def are_order_dates_available(self):
+        capacity_per_day: list[int] = self.get_capacity_of_daycare_in_dates_range(self.daycare_id,
+                                                                                  self.start_date,
+                                                                                  self.end_date)
+
+        daycare_capacity = DayCare.objects.get(name=self.daycare_id.name).capacity
+        return all(current_capacity < daycare_capacity for current_capacity in capacity_per_day)
+
+    def is_the_order_cancelable(self):
+        return self.status == StatusOptions.Pending or self.status == StatusOptions.Approved
+
+    def is_the_order_approvable(self):
+        return self.status == StatusOptions.Pending and self.are_order_dates_available()
--- a/orders/templates/orders/orders.html
+++ b/orders/templates/orders/orders.html
@@ -1,0 +1,54 @@
+{% extends "main/base_template.html" %}
+{% load static %}
+
+{% block stylesheets %}
+<link href="https://cdn.jsdelivr.net/npm/bootstrap@5.0.2/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-EVSTQN3/azprG1Anm3QDgpJLIm9Nao0Yz1ztcQTwFspd3yD65VohhpuuCOmLASjC" crossorigin="anonymous">
+<link rel="stylesheet" type="text/css" href="{% static 'css/orders.css' %}">
+{% endblock %}
+
+{% block content %}
+    <div class="container">
+        <div class="row">
+            <div class="col-12">
+                <table class="table table-bordered">
+                    <thead>
+                    <tr>
+                        <th scope="col">Name</th>
+                        <th scope="col">Book Date</th>
+                        <th scope="col">Start Date</th>
+                        <th scope="col">End Date</th>
+                        <th scope="col">Status</th>
+                        <th scope="col">Actions</th>
+                    </tr>
+                    </thead>
+
+                    <tbody>
+                    {% for order in orders %}
+                        <tr>
+                            <th scope="row">
+                                {% if user == 'daycare' %}
+                                    {{ order.dog_owner_id }}
+                                {% else %}
+                                    {{ order.daycare_id.name }}
+                                {% endif %}
+                            </th>
+                            <td>{{order.book_date}}</td>
+                            <td>{{order.start_date}}</td>
+                            <td>{{order.end_date}}</td>
+                            <td>{{order.get_order_status}}</td>
+                            <td>
+                                {% if order.is_the_order_approvable and user == 'daycare' %}
+                                    <button type="button" class="btn btn-success">Approve</button>
+                                {% endif %}
+                                {% if order.is_the_order_cancelable %}
+                                    <button type="button" class="btn btn-danger">Cancel</button>
+                                {% endif %}
+                            </td>
+                        </tr>
+                    {% endfor %}
+                    </tbody>
+                </table>
+            </div>
+        </div>
+    </div>
+{% endblock %}
--- a/orders/tests.py
+++ b/orders/tests.py
@@ -52,3 +52,66 @@ class TestOrderModel:
     def test_daycare_id_is_deleted_when_daycare_is_deleted(self, create_order):
         DayCare.objects.get(id=1).delete()
         assert Order.objects.get(id=create_order.id).daycare_id is None
+
+    @pytest.mark.parametrize("dog_owner_1_id, dog_owner_2_id, delta_1, delta_2, capacity, expected_result",
+                             [(1, 2, 3, 5, 1, False),
+                              (2, 5, 4, 3, 10, True),
+                              (5, 4, 6, 7, 15, True),
+                              (4, 3, 4, 2, 1, False)])
+    def test_order_is_approvable_according_to_daycare_capacity(self,
+                                                               dog_owner_1_id: DogOwner,
+                                                               dog_owner_2_id: DogOwner,
+                                                               delta_1: int,
+                                                               delta_2: int,
+                                                               capacity: int,
+                                                               expected_result: bool):
+        daycare = DayCare.create("capacity_email@gmail.com", "CapacityUserName", "CapacityPassword", "CapacityName",
+                                 "Changeable capacity", 100, capacity, "Merkaz", "Tel Aviv", "Capacity 123")
+        dog_owner_1 = DogOwner.objects.get(id=dog_owner_1_id)
+        dog_owner_2 = DogOwner.objects.get(id=dog_owner_2_id)
+
+        order1 = Order.create(start_date=timezone.now(), end_date=timezone.now() + datetime.timedelta(days=delta_1),
+                              daycare_id=daycare, dog_owner_id=dog_owner_1, price_per_day=100)
+        order2 = Order.create(start_date=timezone.now(), end_date=timezone.now() + datetime.timedelta(days=delta_2),
+                              daycare_id=daycare, dog_owner_id=dog_owner_2, price_per_day=100)
+
+        order1.approve_order()
+        assert order2.is_the_order_approvable() == expected_result
+
+    @pytest.mark.parametrize("new_status, expected_result", [(StatusOptions.Pending, True),
+                                                             (StatusOptions.Canceled, False),
+                                                             (StatusOptions.Approved, True),
+                                                             (StatusOptions.OnGoing, False),
+                                                             (StatusOptions.Finished, False)])
+    def test_order_cancellation_according_to_order_status(self,
+                                                          create_order: Order,
+                                                          new_status: StatusOptions,
+                                                          expected_result: StatusOptions):
+        create_order.status = new_status
+        assert create_order.is_the_order_cancelable() == expected_result
+
+    @pytest.mark.parametrize("new_status, expected_result", [(StatusOptions.Pending, "Pending"),
+                                                             (StatusOptions.Canceled, "Canceled"),
+                                                             (StatusOptions.Approved, "Approved"),
+                                                             (StatusOptions.OnGoing, "Ongoing"),
+                                                             (StatusOptions.Finished, "Finished")])
+    def test_get_order_status_as_string(self,
+                                        create_order: Order,
+                                        new_status: StatusOptions,
+                                        expected_result: StatusOptions):
+        create_order.status = new_status
+        assert create_order.get_order_status() == expected_result
+
+    def test_get_daycare_capacity_in_dates_range(self, create_order):
+        original_capacity_list = Order.get_capacity_of_daycare_in_dates_range(create_order.daycare_id,
+                                                                              create_order.start_date,
+                                                                              create_order.end_date)
+        create_order.approve_order()
+        updated_capacity_list = Order.get_capacity_of_daycare_in_dates_range(create_order.daycare_id,
+                                                                             create_order.start_date,
+                                                                             create_order.end_date)
+        assert len(original_capacity_list) == len(updated_capacity_list)
+
+        equality_list = [updated_capacity_list[i] == original_capacity_list[i] + 1
+                         for i in range(len(original_capacity_list))]
+        assert equality_list.count(True) == len(equality_list)
--- a/orders/views.py
+++ b/orders/views.py
@@ -1,3 +1,21 @@
-# from django.shortcuts import render
+from django.contrib.auth.decorators import login_required
+from django.shortcuts import render
 
-# Create your views here.
+from dogowner.models import DogOwner
+from orders.models import Order
+
+
+@login_required()
+def orders(request):
+    if DogOwner.objects.filter(user=request.user).exists():
+        context = {
+            'orders': Order.objects.filter(dog_owner_id__user=request.user),
+            'user': 'dog_owner',
+        }
+    else:
+        context = {
+            'orders': Order.objects.filter(daycare_id__user=request.user),
+            'user': 'daycare',
+        }
+
+    return render(request, 'orders/orders.html', context)
--- a/Hotails/urls.py
+++ b/Hotails/urls.py
@@ -15,6 +15,8 @@ Including another URLconf
 """
 from django.contrib import admin
 from django.urls import path
+
+import orders.views
 from main import views
 from django.contrib.auth import views as auth_views
 
@@ -26,4 +28,5 @@ urlpatterns = [
     path('homepage/', views.homepage, name='homepage'),
     path('logout/', views.logout_view, name='logout'),
     path('about/', views.about, name='about'),
+    path('orders/', orders.views.orders, name='orders'),
 ]
--- a/main/templates/main/base_template.html
+++ b/main/templates/main/base_template.html
@@ -25,7 +25,7 @@
             <ul class="navbar-nav mr-auto">
                 {% if request.user.is_authenticated %}
                     <li class="nav-item"><a class="nav-link text-white" href="/profile">Profile</a></li>
-                    <li class="nav-item"><a class="nav-link text-white" href="">Orders</a></li>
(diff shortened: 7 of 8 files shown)


main/views.py
```diff
@@ -3,6 +3,8 @@
 from django.shortcuts import redirect
 from django.contrib.auth import logout
```
ErezCohenn, line 5: a blank line here is unnecessary

orders/models.py
```diff
@@ -71,3 +71,24 @@ def cancel_order(self):
...
+            return "Approved"
+        elif self.status == StatusOptions.Canceled:
+            return "Canceled"
+        elif self.status == StatusOptions.OnGoing:
+            return "On going"
+        return "Finished"
+
+    def is_the_order_cancelable(self):
+        if self.status != StatusOptions.OnGoing or self.status != StatusOptions.Finished:
+            return True
```
ErezCohenn, line 88: I think you should also return true in the status `Pending`  if the user wants to cancel the order.

orders/models.py
```diff
@@ -71,3 +71,24 @@ def cancel_order(self):
...
+        return False
+
+    def is_the_order_approvable(self):
+        if self.status != StatusOptions.Pending:
+            return True
+        return False
```
ErezCohenn, line 94: this method is only for the daycare user, right? I think you also need to check the capacity of the dates of the order and check if there is a place for more dogs on those dates.

orders/templates/orders/orders.html
```diff
@@ -0,0 +1,54 @@
...
+                                {% endif %}
+                            </th>
+                            <td>{{order.book_date}}</td>
+                            <td>{{order.start_date}}</td>
+                            <td>{{order.end_date}}</td>
+                            <td>{{order.get_order_status}}</td>
+                            <td>
+                                {% if order.is_the_order_approvable %}
+                                    <button type="button" class="btn btn-success">Approve</button>
+                                {% endif %}
```
ErezCohenn, line 42: if the method `is_the_order_approvable` is only for daycare as I mentioned, I think you should also check here if the user is a daycare or dog-owner.

ErezCohenn (changes requested):
Great job @OfirMatasas!
take a look at the comment I left.

orders/models.py
OfirMatasas: The option of cancelling the order is when the order is not finished or currently happening.
The pending status is included in that scenario.

main/views.py
OfirMatasas: Pycharm added this line, after reformatting the file. I'll remove it anyway.

main/tests.py
```diff
@@ -1,22 +1,23 @@
...
-                           first_name='test',
-                           last_name='user',
-                           phone_number=1234567890,
-                           dog_race='dog race',
-                           dog_picture_url='https://www.google.com/user1.jpg',
-                           dog_age=4,
-                           dog_weight=2,
-                           dog_gender='M'
-                           )
+    return DogOwner.create(email='dogowner@address.com', username='dogOwnerUser01', password='password123',
```
Omeramsc, line 9: How is this change relevant to this PR?

orders/models.py
```diff
@@ -71,3 +72,56 @@ def cancel_order(self):
...
+
+    def get_order_status(self):
+        if self.status == StatusOptions.Pending:
+            return "Pending"
+        elif self.status == StatusOptions.Approved:
+            return "Approved"
+        elif self.status == StatusOptions.Canceled:
+            return "Canceled"
+        elif self.status == StatusOptions.OnGoing:
+            return "On going"
```
Omeramsc, line 84: s/'On going'/'Ongoing'

orders/models.py
```diff
@@ -71,3 +72,56 @@ def cancel_order(self):
...
+        return "Finished"
+
+    @staticmethod
+    def get_capacity_of_daycare_in_dates_range(daycare_id, start_date, end_date):
+        capacity_per_day_list = [0] * ((end_date - start_date).days + 2)
+        relevant_orders = Order.objects.filter(daycare_id=daycare_id, status__in=['A', 'O'])
+        start_date = datetime.date(year=start_date.year, month=start_date.month, day=start_date.day)
```
Omeramsc, line 91: Are you using the same names on purpose? I'd change them.

orders/models.py
```diff
@@ -71,3 +72,56 @@ def cancel_order(self):
...
+        end_date = datetime.date(year=end_date.year, month=end_date.month, day=end_date.day)
+
+        for order in relevant_orders:
+            if end_date < order.start_date or order.end_date < start_date:
```
Omeramsc, line 95: Can you explain what's the logic here?

orders/models.py
```diff
@@ -71,3 +72,56 @@ def cancel_order(self):
...
+                if current_date < start_date:
+                    continue
+                elif current_date > end_date:
+                    break
+                else:
+                    capacity_per_day_list[day] = capacity_per_day_list[day] + 1
+
+        return capacity_per_day_list
+
+    def are_order_dates_are_not_occupied_for_approval(self):
```
Omeramsc, line 111: The name is a bit confusing, a better one will be: `are_order_dates_available`

orders/models.py
```diff
@@ -71,3 +72,56 @@ def cancel_order(self):
...
+        capacity_per_day_list = self.get_capacity_of_daycare_in_dates_range(self.daycare_id,
```
Omeramsc, line 112: a good example to where using `Typing` will result in a better name and better understood method overall
`capacity_per_day_list: list[int] = self.get_capacity_of_daycare_in_dates_range(...)`

orders/models.py
```diff
@@ -71,3 +72,56 @@ def cancel_order(self):
...
+                                                                            self.start_date,
+                                                                            self.end_date)
+
+        daycare_capacity = DayCare.objects.get(name=self.daycare_id.name).capacity
+        return all(current_capacity < daycare_capacity for current_capacity in capacity_per_day_list)
+
+    def is_the_order_cancelable(self):
+        if self.status == StatusOptions.Pending or self.status == StatusOptions.Approved:
```
Omeramsc, line 120: Instead of the `if`, you can return the result of the conditional.

orders/models.py
```diff
@@ -71,3 +72,56 @@ def cancel_order(self):
...
+            return True
+        return False
+
+    def is_the_order_approvable(self):
+        if self.status == StatusOptions.Pending and self.are_order_dates_are_not_occupied_for_approval():
```
Omeramsc, line 125: same

orders/models.py
```diff
@@ -71,3 +72,56 @@ def cancel_order(self):
...
+        elif self.status == StatusOptions.Approved:
+            return "Approved"
+        elif self.status == StatusOptions.Canceled:
+            return "Canceled"
+        elif self.status == StatusOptions.OnGoing:
+            return "On going"
+        return "Finished"
+
+    @staticmethod
+    def get_capacity_of_daycare_in_dates_range(daycare_id, start_date, end_date):
```
Omeramsc, line 88: missing test

orders/models.py
```diff
@@ -71,3 +72,56 @@ def cancel_order(self):
 
     def get_order_total_price(self):
         return self.price_per_day * (self.end_date - self.start_date).days
+
+    def get_order_status(self):
```
Omeramsc, line 76: missing test

orders/models.py
Omeramsc, line 111: missing test

Omeramsc:
Missing a lot of tests. returning to `in progress`.

main/tests.py
```diff
@@ -87,3 +88,21 @@ def test_unlogged_user_access_to_homepage(self, client):
         response = client.get("/homepage/")
         assert response.status_code == 302
         assert response['Location'] == '/login/?next=/homepage/'
+
+
+@pytest.mark.django_db
+class TestOrdersView:
+    def test_unlogged_user_access_to_orders(self, client):
```
Omeramsc, line 95: I think we can trust Django's `login_required` decorator functionality

main/tests.py
OfirMatasas: It's just shrinks the place this fixture takes. I can revert it it you think it's redundant.

orders/models.py
OfirMatasas: Changed.

orders/models.py
OfirMatasas: Changed "start_date","end_date" to "start_datetime", "end_datetime".

orders/models.py
OfirMatasas: Let's give the orders names for better explanation:
1. Checked order - the order we want to check that can be approved. 
2. Current order - the order from the list of relevant_orders.

We want to check the dates that the current order take place, because if the orders are not having shared dates, then the check for capacity on those dates is irrelevant, and we want to skip that.
So, if the current order is ending before the checked order, or the checked order is ending before the current order, then we can skip to the next order from the list.

orders/models.py
OfirMatasas: Changed to are_order_dates_available.

orders/models.py
OfirMatasas: Changed also on "get_capacity_of_daycare_in_dates_range"

orders/models.py
OfirMatasas: You're right. Changed.

orders/models.py
OfirMatasas: Changed.

main/tests.py
OfirMatasas: Removed.

orders/models.py
OfirMatasas: Added.

orders/models.py
OfirMatasas: Added

orders/models.py
OfirMatasas: I think that the "test_order_is_approvable_according_to_daycare_capacity" covers it, because that's the main purpose for this test.

ErezCohenn (approved):
LGTM !

orders/models.py
Omeramsc: I'm pretty sure you can pull specific dates from the DB instead of dealing with this complicated logic.

main/tests.py
ErezCohenn, line 9: Now that we merged the conftest file, for each there is no need to exists any `@pytest.fixture` code.
in your case you can just use the `createdog_owner_user` in any place you want and you should delete it from here

main/tests.py
```diff
@@ -1,22 +1,23 @@
...
+                           dog_race='dog race', dog_picture_url='https://www.google.com/user1.jpg', dog_age=4,
+                           dog_weight=2, dog_gender='M')
+
+
+@pytest.fixture
+def create_daycare_user():
+    return DayCare.create(email="test@gmail.com", username="testuser01", password="pass",
+                          name="Puppies", description="This is the first daycare test",
+                          price_per_day=10,
+                          capacity=50, area="Merkaz", city="Tel-Aviv", address="The best street 5")
```
ErezCohenn, line 20: same, should delete the function : `create_daycare_user`

ErezCohenn (changes requested):

main/tests.py
OfirMatasas: Update - after the conftest file, this is no longer relevant. Deleted.

OfirMatasas: Removed `createdog_owner_user` fixture

main/tests.py
OfirMatasas: Removed 'create_daycare_user' fixture.

orders/models.py
OfirMatasas: Added ranges of orders to the filter.

ErezCohenn (approved):

main/tests.py
```diff
@@ -69,3 +69,16 @@ def test_unlogged_user_access_to_homepage(self, client):
         response = client.get("/homepage/")
         assert response.status_code == 302
         assert response['Location'] == '/login/?next=/homepage/'
+
+
+@pytest.mark.django_db
+class TestOrdersView:
```
Yuval-Vino, line 75: maybe need to add a test that shows we are displaying only the relevant orders in HTML

Yuval-Vino (commented):
Great job ofir! orders HTML looks good , I left 1 comment tell me what you think

tamirmatok (approved):
Great job @OfirMatasas !!
I left some comments let me know what u think.

main/tests.py
```diff
@@ -77,3 +78,28 @@ def test_dog_owner_homepage_is_visible_for_dog_owner(self, client, create_dog_ow
...
+
+    def test_orders_page_is_visible_for_daycare_user(self, client, create_daycare_user):
+        client.force_login(user=create_daycare_user.user)
+        response = client.get("/orders/")
+        assert response.status_code == 200
+
+    def test_only_relevant_orders_are_displayed_for_daycare(self, client, create_daycare_user):
+        client.force_login(user=create_daycare_user.user)
+        response = client.get("/orders/")
+        assert list(response.context['orders']) == list(Order.objects.filter(daycare_id=create_daycare_user))
```
Yuval-Vino, line 98: Please take a look at https://github.com/redhat-beyond/Hotails/pull/64#discussion_r862467315 about naming variabels in assert

main/tests.py
```diff
@@ -77,3 +78,28 @@ def test_dog_owner_homepage_is_visible_for_dog_owner(self, client, create_dog_ow
...
+        assert response.context['user'] == 'daycare'
+
+    def test_only_relevant_orders_are_displayed_for_dog_owner(self, client, create_dog_owner_user):
+        client.force_login(user=create_dog_owner_user.user)
+        response = client.get("/orders/")
+        assert list(response.context['orders']) == list(Order.objects.filter(dog_owner_id=create_dog_owner_user))
+        assert response.context['user'] == 'dog_owner'
```
Yuval-Vino, line 105: same comment as above

orders/models.py
```diff
@@ -71,3 +72,51 @@ def cancel_order(self):
...
+        elif self.status == StatusOptions.Approved:
+            return "Approved"
+        elif self.status == StatusOptions.Canceled:
+            return "Canceled"
+        elif self.status == StatusOptions.OnGoing:
+            return "Ongoing"
+        return "Finished"
+
+    @staticmethod
+    def get_capacity_of_daycare_in_dates_range(daycare_id, start_datetime, end_datetime):
```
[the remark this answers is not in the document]
Yuval-Vino: I think it means , it is unclear what kind of type you are expecting to get as "start_datetime" and "end_datetime" for instance you can use ` def get_capacity_of_daycare_in_dates_range(daycare_id: int, start_datetime: DateField, end_datetime: DateField):` etc..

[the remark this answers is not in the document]
OfirMatasas: As I see it, daycare_id is pretty explicit as an integer, because that's how the member is defined. Same goes for "start_datetime" and "end_datetime".
Tell me if what do you think.

main/tests.py
OfirMatasas: Changed.

main/tests.py
OfirMatasas: Changed.

main/views.py
```diff
@@ -29,3 +30,7 @@ def about(request):
 def logout_view(request):
     logout(request)
     return index(request)
+
+
+def orders_view(request):
```
Omeramsc, line 35: Is this needed?

main/tests.py
```diff
@@ -77,3 +78,34 @@ def test_dog_owner_homepage_is_visible_for_dog_owner(self, client, create_dog_ow
...
+        client.force_login(user=create_dog_owner_user.user)
+        response = client.get("/orders/")
+        assert response.status_code == 200
+
+    def test_orders_page_is_visible_for_daycare_user(self, client, create_daycare_user):
+        client.force_login(user=create_daycare_user.user)
+        response = client.get("/orders/")
+        assert response.status_code == 200
+
+    def test_only_relevant_orders_are_displayed_for_daycare(self, client, create_daycare_user):
```
Omeramsc, line 95: it doesn't test only relevant orders are displayed. it tests that ALL the orders are displayed...

main/tests.py
```diff
@@ -77,3 +78,34 @@ def test_dog_owner_homepage_is_visible_for_dog_owner(self, client, create_dog_ow
...
+        assert response_user == 'daycare'
+
+    def test_only_relevant_orders_are_displayed_for_dog_owner(self, client, create_dog_owner_user):
+        client.force_login(user=create_dog_owner_user.user)
+        response = client.get("/orders/")
+        response_orders_list = list(response.context['orders'])
+        all_orders_of_dog_owner_list = list(Order.objects.filter(dog_owner_id=create_dog_owner_user))
+        assert response_orders_list == all_orders_of_dog_owner_list
+        response_user = response.context['user']
+        assert response_user == 'dog_owner'
```
Omeramsc, line 111: same as the former test

orders/tests.py
```diff
@@ -52,3 +52,55 @@ def test_dog_owner_id_is_deleted_when_dog_owner_is_deleted(self, create_order):
...
+    @pytest.mark.parametrize("new_status, expected_result", [(StatusOptions.Pending, "Pending"),
+                                                             (StatusOptions.Canceled, "Canceled"),
+                                                             (StatusOptions.Approved, "Approved"),
+                                                             (StatusOptions.OnGoing, "Ongoing"),
+                                                             (StatusOptions.Finished, "Finished")])
+    def test_get_order_status_as_string(self, create_order, new_status, expected_result):
+        create_order.status = new_status
+        assert create_order.get_order_status() == expected_result
+
+    def test_get_daycare_capacity_in_dates_range(self, create_order):
```
Omeramsc, line 94: That's not how you should test this function. it has a lot of logic; you need to test different cases and see that you get the capacity you expect in each case

orders/models.py
```diff
@@ -71,3 +72,51 @@ def cancel_order(self):
...
+                if current_date < start_date:
+                    continue
+                elif current_date > end_date:
+                    break
+                else:
+                    capacity_per_day[day] = capacity_per_day[day] + 1
+
+        return capacity_per_day
+
+    def are_order_dates_available(self):
```
Omeramsc, line 110: test?

Omeramsc (changes requested):

main/tests.py
OfirMatasas: Changed the name of the method to "test_only_orders_which_belongs_to_user_as_daycare_are_displayed". Same goes for dog owner.

main/tests.py
```diff
@@ -77,3 +78,34 @@ def test_dog_owner_homepage_is_visible_for_dog_owner(self, client, create_dog_ow
...
+class TestOrdersView:
+    def test_orders_page_is_visible_for_dog_owner_user(self, client, create_dog_owner_user):
+        client.force_login(user=create_dog_owner_user.user)
+        response = client.get("/orders/")
+        assert response.status_code == 200
+
+    def test_orders_page_is_visible_for_daycare_user(self, client, create_daycare_user):
+        client.force_login(user=create_daycare_user.user)
+        response = client.get("/orders/")
+        assert response.status_code == 200
```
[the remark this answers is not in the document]
OfirMatasas: Removed.

orders/models.py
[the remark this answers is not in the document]
OfirMatasas: Added typing.

main/tests.py
```diff
@@ -77,3 +78,34 @@ def test_dog_owner_homepage_is_visible_for_dog_owner(self, client, create_dog_ow
...
+
+    def test_only_relevant_orders_are_displayed_for_daycare(self, client, create_daycare_user):
+        client.force_login(user=create_daycare_user.user)
+        response = client.get("/orders/")
+        response_orders_list = list(response.context['orders'])
+        all_orders_of_daycare_list = list(Order.objects.filter(daycare_id=create_daycare_user))
+        assert response_orders_list == all_orders_of_daycare_list
+        response_user = response.context['user']
+        assert response_user == 'daycare'
```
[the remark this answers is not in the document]
OfirMatasas: Switched.

main/tests.py
```diff
@@ -77,3 +78,34 @@ def test_dog_owner_homepage_is_visible_for_dog_owner(self, client, create_dog_ow
...
+    def test_orders_page_is_visible_for_daycare_user(self, client, create_daycare_user):
+        client.force_login(user=create_daycare_user.user)
+        response = client.get("/orders/")
+        assert response.status_code == 200
+
+    def test_only_relevant_orders_are_displayed_for_daycare(self, client, create_daycare_user):
+        client.force_login(user=create_daycare_user.user)
+        response = client.get("/orders/")
+        response_orders_list = list(response.context['orders'])
+        all_orders_of_daycare_list = list(Order.objects.filter(daycare_id=create_daycare_user))
```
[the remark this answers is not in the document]
OfirMatasas: Separated.

orders/models.py
OfirMatasas: I made "test_order_is_approvable_according_to_daycare_capacity" for this method. Should I test it in another way?

orders/tests.py
OfirMatasas: I thought that checking the current capacity of a daycare before and after approving an order will check if the method is working correctly.
Also, some of the other tests are testing this method too (like "test_order_is_approvable_according_to_daycare_capacity").

Do you have any example for a case that I should test for this method? Or another way for testing it?

main/tests.py
OfirMatasas: Switched.

orders/models.py
Omeramsc: Is it testing this function alone? 
if it does, it deserves a name change :)

Omeramsc: The only thing this function does is check if specific dates are available due to capacity. the test you wrote deal with the whole process your application does. What we need for this function is a **unit** test, as it tests this unit specifically, without dependencies from other processes.

orders/tests.py
Omeramsc: second:
the test above test if order can be approved, according to its overall calculated capacity
this test should be a specific unit-test for `get_capacity_of_daycare_in_dates_range` it doesn't test the overall logic your app will go through, but it is testing this method specifically, and should pass even if in the future you'll decide to change the way the process works. 

this method has a lot of cases. think of each `if` statement as at least two possible test-cases.

orders/models.py
```diff
@@ -71,3 +72,51 @@ def cancel_order(self):
...
+    def get_order_status(self):
+        if self.status == StatusOptions.Pending:
+            return "Pending"
+        elif self.status == StatusOptions.Approved:
+            return "Approved"
+        elif self.status == StatusOptions.Canceled:
+            return "Canceled"
+        elif self.status == StatusOptions.OnGoing:
+            return "Ongoing"
+        return "Finished"
```
ormergi, line 85: I think this can be simplified to
```
statusOptionsNameIdx=1
reuturn self.status[statusOptionsNameIdx]
```
WDYT?

orders/models.py
```diff
@@ -71,3 +72,51 @@ def cancel_order(self):
...
+
+    @staticmethod
+    def get_capacity_of_daycare_in_dates_range(daycare_id: DayCare, start_datetime: datetime, end_datetime: datetime):
```
ormergi, line 88: What is capacity?

orders/models.py
```diff
@@ -71,3 +72,51 @@ def cancel_order(self):
...
+        number_of_days = (end_datetime - start_datetime).days
+        capacity_per_day: list[int] = [0] * number_of_days
+        start_date = datetime.date(year=start_datetime.year, month=start_datetime.month, day=start_datetime.day)
+        end_date = datetime.date(year=end_datetime.year, month=end_datetime.month, day=end_datetime.day)
+        relevant_orders = Order.objects.filter(daycare_id=daycare_id,
+                                               status__in=['A', 'O'],
```
ormergi, line 94: Is it possible to use full names?

orders/tests.py
ormergi: @OfirMatasas are you going to address these?

orders/views.py
```diff
@@ -1,3 +1,21 @@
...
-# Create your views here.
+from dogowner.models import DogOwner
+from orders.models import Order
+
+
+@login_required()
+def orders(request):
+    if DogOwner.objects.filter(user=request.user).exists():
+        context = {
+            'orders': Order.objects.filter(dog_owner_id__user=request.user),
```
ormergi, line 12: Why is it necessary to fillter by authenticated user, is it possible to use `request.user` directly?

ormergi (commented):
Hi Ofir, please see my inline comments

Omeramsc:
There are conflicts and a bunch of unresolved comments. Sorry, it won't be ready for the demo.

PR #97, opened 2022-04-30
Create static orders

### LINKED ISSUE #96: Add static orders
## Expected Behavior

Because the project uses only static users without the ability to register at the moment, there is a need for a large number of static orders for regular work.

##Current Behavior

There are only 4 static orders in the data migrations.

## Acceptance Criteria

Add static orders for all users, when they cover all possible statuses of orders.

OfirMatasas:
### Desired Outcome

Orders are created automatically for daycares and dog owners

### Implemented Changes

New migration that create orders for daycares and dog owners

### Connected Issue/Story

Closes #96 

#### Changelog

  CHANGELOG update

#### Test coverage

- [ ] This PR includes new unit and integration tests to go with the code
  changes
- [x] The changes in this PR do not require tests

#### Documentation


orders/migrations/0003_static_orders.py
```diff
@@ -0,0 +1,79 @@
+from django.db import migrations, transaction
+from django.utils import timezone
+
+DOG_OWNER_ID = ['4', '5', '6', '7', '8', '9', '10', '11', '12', '13', '14', '15', '16', '17', '18', '19', '20', '21',
+                '22', '23', '4', '5', '6', '7', '8', '9', '10', '11', '12', '13', '14', '15', '16', '17', '18', '19',
+                '20', '21', '22', '23']
+
+DAYCARE_ID = ['4', '5', '6', '7', '8', '9', '10', '11', '12', '13', '14', '15', '16', '17', '18', '19', '20', '21',
```
Yuval-Vino, line 8: What's going on here, what are those id`s?

Yuval-Vino (commented):
I don't think this PR should include pipfile changes?

orders/migrations/0003_static_orders.py
```diff
@@ -0,0 +1,79 @@
...
+              timezone.datetime(2022, 7, 8), timezone.datetime(2022, 8, 1), timezone.datetime(2022, 1, 1),
+              timezone.datetime(2022, 7, 7), timezone.datetime(2022, 7, 20), timezone.datetime(2022, 8, 27),
+              timezone.datetime(2022, 1, 15), timezone.datetime(2022, 12, 18), timezone.datetime(2022, 9, 23),
+              timezone.datetime(2022, 11, 12), timezone.datetime(2022, 11, 11), timezone.datetime(2022, 1, 23),
+              timezone.datetime(2022, 6, 12), timezone.datetime(2022, 2, 15), timezone.datetime(2022, 3, 11),
+              timezone.datetime(2022, 1, 17), timezone.datetime(2022, 12, 23), timezone.datetime(2022, 3, 12),
+              timezone.datetime(2022, 3, 5), timezone.datetime(2022, 4, 4), timezone.datetime(2022, 5, 15),
+              timezone.datetime(2022, 5, 5)]
+
+END_DATE = [timezone.datetime(2022, 3, 30), timezone.datetime(2022, 5, 25), timezone.datetime(2022, 3, 1),
```
Yuval-Vino, line 27: maybe instead of keeping the END_DATE list, use START_DATE + random number of days (in our order days range)

OfirMatasas:
> I don't think this PR should include pipfile changes?

You're right. Something went wrong there. Removing it now.

orders/migrations/0003_static_orders.py
OfirMatasas: Those are the id's of the static daycares and the static dog owners, from previous pull requests.
I used them to populate Hotails with orders from existed users.

orders/migrations/0003_static_orders.py
```diff
@@ -0,0 +1,79 @@
...
+            timezone.datetime(2022, 7, 14), timezone.datetime(2022, 8, 19), timezone.datetime(2022, 1, 17),
+            timezone.datetime(2022, 7, 17), timezone.datetime(2022, 7, 25), timezone.datetime(2022, 9, 5),
+            timezone.datetime(2022, 1, 24), timezone.datetime(2022, 12, 24), timezone.datetime(2022, 10, 1),
+            timezone.datetime(2022, 11, 24), timezone.datetime(2022, 11, 22), timezone.datetime(2022, 2, 3),
+            timezone.datetime(2022, 6, 20), timezone.datetime(2022, 2, 28), timezone.datetime(2022, 3, 25),
+            timezone.datetime(2022, 1, 22), timezone.datetime(2022, 12, 26), timezone.datetime(2022, 3, 26),
+            timezone.datetime(2022, 3, 17), timezone.datetime(2022, 4, 17), timezone.datetime(2022, 5, 26),
+            timezone.datetime(2022, 5, 14)]
+
+PRICE_PER_DAY = [200, 230, 150, 60, 160, 300, 90, 150, 170, 50, 200, 230, 150, 60, 160, 300, 90, 150, 170, 50, 200, 230,
```
Yuval-Vino, line 42: why not use the daycare field price per day?

orders/migrations/0003_static_orders.py
OfirMatasas: It can be done, but what's the benefit of using it?

orders/migrations/0003_static_orders.py
OfirMatasas: You're right. Good one.

orders/migrations/0003_static_orders.py
Yuval-Vino: maybe I'm wrong, but for me, it looks more elegant than the (not so pretty) END_DATE list

orders/migrations/0003_static_orders.py
Yuval-Vino: maybe for daycare you can try this   static_daycares = [daycare for daycare in DayCare.objects.all() if 'static' in daycare.user.username]
you then concat it with static_daycares[::-1]
** not so sure about the static dogonwers

OfirMatasas: I'm not sure if depending on their name is better than depending on their id, because they're static, and they should remain as they are ATM.

Yuval-Vino: Maybe so, yet this is not so readable because we won't remember the ids, but if we call all the "static daycares" it will be clear for everyone

OfirMatasas: After checking in the database about the static daycares, I see that they don't have the word "static" in their name, or anything else. I'd change it to the way you suggested for better readability, but apparently it can't be done.

image

OfirMatasas: Ok You're right. I'll change it.

OfirMatasas: After trying to change it to the way you suggested, a problem occurred - because the migration of the daycares is in the same process of the migration of the orders, it can't be done.
Plus, the dog owners don't have any identifier like the daycares, so to be consistent, I'll keep the array of IDs for both daycares and dog owners.

orders/migrations/0003_static_orders.py
```diff
@@ -0,0 +1,69 @@
...
+              timezone.datetime(2022, 1, 17), timezone.datetime(2022, 12, 23), timezone.datetime(2022, 3, 12),
+              timezone.datetime(2022, 3, 5), timezone.datetime(2022, 4, 4), timezone.datetime(2022, 5, 15),
+              timezone.datetime(2022, 5, 5)]
+
+END_DATE_DELTA = [7, 2, 8, 3, 6, 2, 7, 2, 1, 5,
+                  15, 4, 11, 3, 9, 2, 5, 7, 3, 4,
+                  10, 14, 15, 3, 8, 6, 7, 6, 4, 5,
+                  9, 7, 5, 2, 3, 17, 8, 2, 1, 3]
+
+
```
ErezCohenn, line 34: Good Job @OfirMatasas 
In general,  wherever you have created a list of values I think you should insert the values in a loop or use a list comprehension instead of the existing situation.

orders/migrations/0003_static_orders.py
```diff
@@ -0,0 +1,69 @@
...
+DOG_OWNER_ID = ['4', '5', '6', '7', '8', '9', '10', '11', '12', '13',
+                '14', '15', '16', '17', '18', '19', '20', '21', '22', '23',
+                '4', '5', '6', '7', '8', '9', '10', '11', '12', '13',
+                '14', '15', '16', '17', '18', '19', '20', '21', '22', '23']
+
+DAYCARE_ID = ['4', '5', '6', '7', '8', '9', '10', '11', '12', '13', '14', '15', '16', '17', '18', '19', '20', '21',
+              '22', '23', '23', '22', '21', '20', '19', '18', '17', '16', '15', '14', '13', '12', '11', '10', '9', '8',
+              '7', '6', '5', '4']
+
+START_DATE = [timezone.datetime(2022, 3, 23), timezone.datetime(2022, 5, 23), timezone.datetime(2022, 2, 23),
```
jeniaSakirko, line 14: I think having it in tuple with the end date will be more readable for example:
`[(timezone.datetime(2022, 3, 23), 7), (timezone.datetime(2022, 5, 23),2)]`

orders/migrations/0003_static_orders.py
```diff
@@ -0,0 +1,69 @@
+from django.db import migrations, transaction
+from django.utils import timezone
+import datetime
+
+DOG_OWNER_ID = ['4', '5', '6', '7', '8', '9', '10', '11', '12', '13',
+                '14', '15', '16', '17', '18', '19', '20', '21', '22', '23',
+                '4', '5', '6', '7', '8', '9', '10', '11', '12', '13',
+                '14', '15', '16', '17', '18', '19', '20', '21', '22', '23']
```
jeniaSakirko, line 8: you have duplicates of the id's

jeniaSakirko (changes requested):

orders/migrations/0003_static_orders.py
OfirMatasas: Great idea! Changed.

orders/migrations/0003_static_orders.py
OfirMatasas: I wanted the dog owners to have 2 orders each, so I had to duplicate the IDs of the dog owners.
I've changed it to [1, 2, 3...] * 2 for better readability.

orders/migrations/0003_static_orders.py
OfirMatasas: Good idea.
Anyway, I've changed the code to Jenia's suggestion, so it's no longer relevant.

orders/migrations/0003_static_orders.py
ErezCohenn: I agree with @Yuval-Vino,
about the dog-owner id I don't see that there is common information for any field like the 'static' prefix in daycare, so maybe run in a loop instead in the range of 4 to 23, maybe something like that:
`DOG_OWNER_ID = [id for id in range(4,24)]`

jeniaSakirko (approved):
LGTM, I was convinced in about the DAYCARE_ID list
@OfirMatasas don't forget to send about the PR being 'simple' in slack

jeniaSakirko (approved):

OfirMatasas:
> LGTM, I was convinced in about the DAYCARE_ID list @OfirMatasas don't forget to send about the PR being 'simple' in slack

Will do, first thing in the morning!

orders/migrations/0003_static_orders.py
OfirMatasas: I've changed the implementation of the lists according to your suggestions.
Thanks for the help guys! appreciate it :)

ErezCohenn (approved):
LGTM !

ErezCohenn (approved):

jeniaSakirko (approved):
LGTM

Yuval-Vino (approved):

Yuval-Vino:
LGTM good work!

tamirmatok (approved):
LGTM

orders/migrations/0003_static_orders.py
```diff
@@ -0,0 +1,154 @@
...
+                                 price_per_day=200, start_date=timezone.datetime(2022, 5, 5),
+                                 end_date=timezone.datetime(2022, 5, 5) + datetime.timedelta(days=8))
+        ]
+
+        approved_orders_indexes = [1, 4, 7, 10, 14, 16, 17, 20, 24, 27, 30, 32, 35, 36, 39]
+        canceled_orders_indexes = [2, 3, 12, 18, 22, 29, 31, 38]
+        with transaction.atomic():
+            for i in approved_orders_indexes:
+                static_orders_list[i].approve_order()
+            for i in canceled_orders_indexes:
```
Omeramsc, line 149: is all this needed? why don't you create them as approved/canceled order?

OfirMatasas: In the "create" method on the "Order" model, when an order is created, it's always on "Pending" status, because the dog owner is waiting for the daycare to approve/decline the order. So after creating an order, the "approve_order()"/ "cancel_order()" methods must be called.
I can call the methods right after the creation of each of the orders for better readability. 
Let me know what you think.

ormergi: Please change it in away that there is no need to keep the indexes of each type of `order`.
I would be nice if there were helpers for creating approved /  canceled orders.

ormergi (changes requested):

orders/migrations/0003_static_orders.py
Omeramsc: When creating the static data you don't need to follow how it'll work in real-time, you just insert the data the way you want it to be. if you are doing more operations on it, it is not that "static" anymore. 
You can create the objects as approved/cancelled, no need for extra steps.

OfirMatasas: I removed the indexes for each order's approval/cancellation, and add to order's "create" method a default parameter of "StatusOptions.Pending".

BUT, for better readability, and explicit creation of a static order, with the desired status, I've created orders with "StatusOptions.Pending" status on the migration file.

Omeramsc: much MUCH better! good job

Omeramsc:
LGTM

